PR #63, opened 2020-08-05
fix tooltip showing when element is deleted

### LINKED ISSUE #40: Fast navigation causes tool-tip glitch
Navigating too quickly causes the item tooltip to hide in the top corner, it doesn't update x/y quick enough, after a split second it does fix though, so it's not urgent.

image

Nitwel:
Fixes #40


benhaynes:
Thank you!!!

PR #85, opened 2020-08-08
Allow changing public path with env variables

WoLfulus:
This allows us to build a properly configured version of app within the PHP api context. Also fixes some 404s on sub paths.

### DIFF
--- a/app/public/index.html
+++ b/app/public/index.html
@@ -13,6 +13,9 @@
       <strong>We're sorry but Directus doesn't work without JavaScript enabled. Please enable it to continue.</strong>
     </noscript>
     <div id="app"></div>
+    <script type="text/javascript">
+      window.$directusAssetBasePath = "<%= BASE_URL %>";
+    </script>
     <!-- built files will be auto injected -->
   </body>
 </html>
--- a/app/public/index.html
+++ b/app/public/index.html
@@ -4,8 +4,8 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover">
-    <link rel="shortcut icon" href="/favicon.ico">
-    <link rel="manifest" href="/manifest.webmanifest" />
+    <link rel="shortcut icon" href="<%= BASE_URL %>favicon.ico">
+    <link rel="manifest" href="<%= BASE_URL %>manifest.webmanifest" />
     <title>Directus</title>
   </head>
   <body class="auto">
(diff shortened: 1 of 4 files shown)


rijkvanzanten:
What are you trying to achieve exactly? 

This public path flag is only used in buildtime and thus won't enable paths support in actual use (also, you missed a couple other places where this public path is relied on as well, like the router)

WoLfulus:
For the installation process on php side, I'm fetching the release from GitHub and building it with the configured mount path on the project. I'm passing that path into the env variable when invoking the build script. 

For the paths I missed, I'll take a look and submit some more changes

rijkvanzanten:
Building it at install time also means that you're now adding Node 10+ as one of the requirements of the PHP API, doesn't sound ideal.. Why not bundle it in build-time?

WoLfulus:
I really would prefer app to support being mounted to any subpath, but as we mentioned on Discord, we have no way to that right now, so I'd fall to building it for a while.

Assuming that by build-time you mean that we bundle an app release whenever we make a PHP release:

Packagist works differently than npm, and I haven't found a (comfortable) way to bundle built asset files into php packages other than commiting them to the repository itself, and I found myself commiting 200+ files for some minor changes when tweaking the GraphQL app ui. While we can still do this, it comes with some problems whenever you want to make some changes on how you expose Directus since the public folder is controlled by the project holding Directus as a dependency (there's no `.static(path)` in Laravel unfortunately like node), user should be able to ask whenever he wants something published to that folder (vendor:publish and/or additional artisan commands) 

The way I'm picturing this is:

- Installation through `composer global require directus/installer && directus new project`
  > This creates a project structure known by Directus, and since it's new we can publish files automatically
  - Installer can fetch the matching (built) release from GitHub Releases and puts it into `./public/admin` folder.
    > Assuming we'll make .zip releases for `directus/next` repository, otherwise we still need to build it
- Installation through `composer require directus/directus`
  > No way to know or infer the project structure, publishing of files should be "asked" by the developer
  - Installs Directus as a package into a Laravel application
    > Doesn't publish files by default - this is package's behavior since Laravel has vendor:publish for publishing files (https://laravel.com/docs/7.x/packages#public-assets)
  - Can use `php artisan directus:admin:install` to fetch/build the ui with path matching route configs
  - Can use `php artisan directus:playgrdound:install` to fetch/build the ui with path matching route configs
- Installation through .zip
  - Already bundles a built version of app + php dependencies - also a known structure

Also, Laravel itself uses node/npm/webpack when using mix, so we are not adding anything new in that case.

https://laravel.com/docs/7.x/mix#installation

rijkvanzanten:
I'll close this for now while we figure out a dynamic way of rendering the assets path in the app. We'll most likely end up adding an `index.php` file to the app that can dynamically set the "real" public URL in runtime instead of build time (as we won't be making a build "on demand" per user)

PR #134, opened 2020-08-15
Add markdown interface

Nitwel:
GIF 15 08 2020 12-21-30

### DIFF
--- a/app/src/interfaces/markdown/markdown.vue
+++ b/app/src/interfaces/markdown/markdown.vue
@@ -49,11 +49,11 @@ export default defineComponent({
 		},
 	},
 	setup(props) {
-		const currentTab = ref(0);
+		const currentTab = ref([0]);
 
-		const html = computed(() => marked(props.value));
-		const showEdit = computed(() => !props.tabbed || currentTab.value === 0);
-		const showPreview = computed(() => !props.tabbed || currentTab.value !== 0);
+		const html = computed(() => (props.value ? marked(props.value) : ''));
+		const showEdit = computed(() => !props.tabbed || currentTab.value[0] === 0);
+		const showPreview = computed(() => !props.tabbed || currentTab.value[0] !== 0);
 
 		return { html, currentTab, showEdit, showPreview };
 	},
--- a/app/src/interfaces/markdown/markdown.vue
+++ b/app/src/interfaces/markdown/markdown.vue
@@ -1,25 +1,25 @@
 <template>
-	<div class="v-markdown" :class="{ tabbed }">
+	<div class="interface-markdown" :class="{ tabbed }">
 		<div v-if="tabbed" class="toolbar">
-			<v-tabs v-model="selection">
+			<v-tabs v-model="currentTab">
 				<v-tab>
 					<v-icon name="code" left />
-					Edit
+					{{ $t('edit') }}
 				</v-tab>
 				<v-tab>
 					<v-icon name="visibility" left />
-					Preview
+					{{ $t('preview') }}
 				</v-tab>
 			</v-tabs>
 		</div>
 		<v-textarea
-			v-show="!tabbed || selection == 0"
+			v-show="!tabbed || currentTab == 0"
 			:placeholder="placeholder"
 			:value="value"
 			:disabled="disabled"
 			@input="$listeners.input"
 		/>
-		<div v-show="!tabbed || selection != 0" class="preview-container">
+		<div v-show="!tabbed || currentTab != 0" class="preview-container">
 			<div class="preview" v-html="html"></div>
 		</div>
 	</div>
@@ -49,17 +49,17 @@ export default defineComponent({
 		},
 	},
 	setup(props) {
-		const selection = ref(0);
+		const currentTab = ref(0);
 
 		const html = computed(() => marked(props.value));
 
-		return { html, selection };
+		return { html, currentTab };
 	},
 });
 </script>
 
 <style lang="scss" scoped>
-.v-markdown {
+.interface-markdown {
 	--v-textarea-min-height: 200px;
 	--v-textarea-max-height: 400px;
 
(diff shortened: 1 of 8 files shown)


app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	&.tabbed .preview-container {
+		border-top: none;
+		border-radius: 0 0 var(--border-radius) var(--border-radius);
+	}
+}
+</style>
+
+<style lang="scss">
```
rijkvanzanten, line 110: Can we make this a v-deep inside the scoped block instead of these global styles?

app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
...
+.v-markdown {
+	.preview {
```
rijkvanzanten, line 112: I believe we have a similar set of styles somewhere for the wysiwyg editor. Would be nice if we could reuse those

app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
...
+		const html = computed(() => marked(props.value));
+
+		return { html, selection };
+	},
+});
+</script>
+
+<style lang="scss" scoped>
+.v-markdown {
+	--v-textarea-min-height: 200px;
```
rijkvanzanten, line 63: Let's make the min height `var(--input-tall)` so it aligns nicely with all others tall interfaces

app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
...
+			type: String,
+			default: null,
+		},
+		tabbed: {
+			type: Boolean,
+			default: true,
+		},
+	},
+	setup(props) {
+		const selection = ref(0);
```
rijkvanzanten, line 52: Let's rename this "currentTab" so it's a little clearer what it is. I thought it was the text selection at first :)

app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
...
+
+		return { html, selection };
+	},
+});
+</script>
+
+<style lang="scss" scoped>
+.v-markdown {
+	--v-textarea-min-height: 200px;
+	--v-textarea-max-height: 400px;
```
rijkvanzanten, line 64: Should we make the editor part monospace? @benhaynes

app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
...
+			</v-tabs>
+		</div>
+		<v-textarea
+			v-show="!tabbed || selection == 0"
+			:placeholder="placeholder"
+			:value="value"
+			:disabled="disabled"
+			@input="$listeners.input"
+		/>
+		<div v-show="!tabbed || selection != 0" class="preview-container">
```
rijkvanzanten, line 22: Let's make this a little more verbose so it's easier to read, for example:

`v-show="tabbed && selection === 1"`

rijkvanzanten: My example is wrong, which goes to show that the original is too hard to read hahah 😬

rijkvanzanten: We can also move them into setup in a computed prop if the inline check gets too tricky

app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
+<template>
+	<div class="v-markdown" :class="{ tabbed }">
+		<div v-if="tabbed" class="toolbar">
+			<v-tabs v-model="selection">
+				<v-tab>
+					<v-icon name="code" left />
+					Edit
```
rijkvanzanten, line 7: Use translations for these pieces of text

app/src/interfaces/markdown/markdown.vue
```diff
@@ -0,0 +1,343 @@
+<template>
+	<div class="v-markdown" :class="{ tabbed }">
```
rijkvanzanten, line 2: Let's use interface-markdown as the class. This isn't a v- base component

app/src/interfaces/markdown/markdown.vue
Nitwel: I'm not sure why but it gives me a i18n is undefined in storybook here.

app/src/interfaces/markdown/markdown.vue
Nitwel: You sure this will not overengineer things?

app/src/interfaces/markdown/markdown.vue
Nitwel: I'll leave the 200px in at the moment because less than 200px kinda screws up the usability.

app/src/interfaces/markdown/markdown.vue
rijkvanzanten: Overengineering is is the act of designing a product to be more robust or have more features than often necessary for its intended use<sup>[1]</sup>. In this case, we'll only write it out a little more verbose in order to make it easier to maintain in the long term 🙂 

<sup>[1]</sup> https://en.wikipedia.org/wiki/Overengineering

app/src/interfaces/markdown/markdown.vue
benhaynes: Hmm, I can mess with that once this is merged in... that's just fine polish.

app/src/interfaces/markdown/markdown.vue
rijkvanzanten: I think you'll have to import it separately (`import i18n from '@/lang'`) and register it in the component: 

```
... defineComponent({ 
  i18n,
...
```

app/src/interfaces/markdown/markdown.vue
rijkvanzanten: No lets set it to `var(--input-height-tall)` regardless so it's consistent with all other tall interfaces. Otherwise this one will stand out like a sore thumb in the form next to the other interfaces. We can always increase what the size of `var(--input-height-tall)` is if it turns out that all the tall interfaces need a little more space 🙂

Nitwel: Ah okey, got it.

PR #139, opened 2020-08-17
Add rate limiter

### LINKED ISSUE #78: Add rate limiter

kukulaka:
Basic rate limiting for all routes. 

Closes #78


kukulaka:
Going to add redis settings to cli

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules
 .env
 npm-debug.log
 lerna-debug.log
+api/example.env
```
rijkvanzanten, line 7: The example env should be tracked in Git

api/package.json
```diff
@@ -52,9 +52,9 @@
   },
   "scripts": {
     "start": "cross-env NODE_ENV=production node dist/server.js",
-    "build": "rm -rf dist && tsc -b && copyfiles \"src/**/*.*\" -e \"src/**/*.ts\" -u 1 dist",
-    "dev": "cross-env NODE_ENV=development LOG_LEVEL=trace ts-node-dev --files src/server.ts --respawn --watch \"src/**/*.ts\" --transpile-only",
-    "cli": "cross-env NODE_ENV=development ts-node --script-mode --transpile-only src/cli/index.ts",
+    "build": "rmdir dist /s && tsc -b && copy \"src/**/*.*\" -e \"src/**/*.ts\" -u 1 dist",
+    "dev": "cross-env  NODE_ENV=development LOG_LEVEL=trace ts-node-dev --files src\\server.ts --respawn --watch \"src/**/*.ts\" --transpile-only",
+    "cli": "cross-env NODE_ENV=development ts-node --script-mode --transpile-only src\\cli\\index.ts",
```
rijkvanzanten, line 57: Does this work cross-platform? (Never realized the differences between Win/Mac were so annoying to deal with 🙄 )

api/package.json
```diff
@@ -85,6 +85,7 @@
     "express": "^4.17.1",
     "express-async-handler": "^1.1.4",
     "express-pino-logger": "^5.0.0",
+    "express-rate-limit": "^5.1.3",
```
rijkvanzanten, line 88: This doesn't seem to be used anywhere

api/package.json
```diff
@@ -104,6 +105,8 @@
     "otplib": "^12.0.1",
     "pino": "^6.4.1",
     "pino-colada": "^2.1.0",
+    "rate-limiter-flexible": "^2.1.10",
+    "redis": "^3.0.2",
```
rijkvanzanten, line 109: Lets make this an optional dependency

api/src/exceptions/redis-not-found.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseException } from './base';
+
+export class RedisNotFoundException extends BaseException {
+	constructor(message: string) {
+		super(message, 404, 'REDIS_NOT_FOUND');
```
rijkvanzanten, line 5: "Redis not found" sounds like a configuration option, and should therefore return a 503 Service Unavailable 

(Also, we should do the same for a wrong MySQL configuration. We can take a peak at [the exceptions that are in the PHP API](https://github.com/directus/api-next/tree/master/src/Support/Exceptions) and lift those over, but lets do that in a separate PR)

api/src/middleware/rate-limiter.ts
```diff
@@ -0,0 +1,64 @@
...
+import env from '../env';
+
+const redisClient = redis.createClient({
+	enable_offline_queue: false,
+	host: env.REDIS_HOST,
+	port: env.REDIS_PORT,
+	password: env.REDIS_PASSWORD,
+});
+
+const rateLimiter: RequestHandler = (req, res, next) => {
```
rijkvanzanten, line 21: Change this to an async/await function:

```suggestion
const rateLimiter: RequestHandler = asyncHandler(async (req, res, next) => {
```

rijkvanzanten: (async handler comes from `import asyncHandler from 'express-async-handler'`)

api/src/middleware/rate-limiter.ts
```diff
@@ -0,0 +1,64 @@
...
+		};
+
+		const rateLimiterRedis = new RateLimiterRedis(opts);
+
+		rateLimiterRedis
+			.consume(req.ip)
+			.then((rateLimiterRes) => {
+				// everything is ok - can put addition logic in there later for users etc
+				next();
+			})
```
rijkvanzanten, line 47: Then with async above, we can change this block to

```js
try {
  const rateLimiterRes = await rateLimiterRedis.consume(req.ip);
catch (err) {
  // etc
}
```

instead of having the double try/catch

api/src/middleware/rate-limiter.ts
kukulaka: Sounds good to me 👍

api/package.json
kukulaka: Need to check this - I suspect not :/

kukulaka: Hiya, so I've put some scripts specifically for windows. When your using command-line rather than bash some of the commands are a bit different, especially around deleting files/directories.

kukulaka:
Have added a fall over so if redis fails for whatever reason  rate limiting  managed in current process memory as per 

https://github.com/animir/node-rate-limiter-flexible/wiki/Memory

api/src/middleware/rate-limiter.ts
```diff
@@ -20,33 +20,38 @@ const redisClient = redis.createClient({
...
+		const redisOpts = {
+			...opts,
+			storeClient: redisClient,
+			// Custom
+			execEvenly: env.EXEC_EVENLY, // delay actions after first action - this may need adjusting (leaky bucket)
+			blockDuration: env.BLOCK_POINT_DURATION, // Do not block if consumed more than points
+			inmemoryBlockOnConsumed: env.INMEMORY_BLOCK_CONSUMED, // eg if 200 points consumed
+			inmemoryBlockDuration: env.INMEMEMORY_BLOCK_DURATION, // block for certain amount of seconds
+		};
+
```
rijkvanzanten, line 44: Can we read / transform these options dynamically, so we don't have to map it manually like this? The storage provider does the same thing:

https://github.com/directus/next/blob/5b1ead8000a132ac6016d93a7d152161320aa496/api/src/storage.ts#L21-L44

That allows us to pass the env vars 1-1 to the options of the library, meaning that we always offer full support for whatever the library offers, even if the library upgrades and adds more options down the line

api/src/utils/get-rate-limiter-config.ts
```diff
@@ -0,0 +1,85 @@
...
+		}
+		if (key === 'INMEMORY_BLOCK_CONSUMED') {
+			redisConfig.inmemoryBlockOnConsumed = value;
+			continue;
+		}
+		if (key === 'INMEMEMORY_BLOCK_DURATION') {
+			redisConfig.inmemoryBlockDuration = value;
+			continue;
+		}
+	}
```
rijkvanzanten, line 80: Can we set the redisConfig dynamically based on the camel-cased value of the env var?

For example `REDIS_POINTS` => `redisConfig.points`, `REDIS_INMEMORY_BLOCK_ON_CONSUMED` => `redisConfig.inmemoryBlockOnConsumed`

That way we always support every option of the Redis/Rate limiting library, regardless of version / used adapter :)

kukulaka: Good idea. Think that's what you meant before, sorry about that. Need more coffee, clearly :D

rijkvanzanten: That is what I meant before yeah 😉 ☕

kukulaka: Done :) So glad don't have any nested options

api/src/utils/get-rate-limiter-config.ts
```diff
@@ -0,0 +1,67 @@
...
+		host: env.SREDIS_HOST,
+		port: env.SREDIS_PORT,
+		password: env.SREDIS_PASSWORD,
+	});
+
+	if (!redisClient) {
+		throw new RedisNotFoundException('Redis client does not exist');
+	}
+
+	redisConfig.keyPrefix = 'rlflx';
```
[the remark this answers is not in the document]
kukulaka: this is just a default but we don't actually need it.

api/src/middleware/rate-limiter.ts
```diff
@@ -0,0 +1,25 @@
...
+const rateLimiter: RequestHandler = asyncHandler(async (req, res, next) => {
+	try {
+		await rateLimiterConfig.consume(req.ip);
+	} catch (rejRes) {
+		// If there is no error, rateLimiterRedis promise rejected with number of ms before next request allowed
+		const secs = Math.round(rejRes.msBeforeNext / 1000) || 1;
+		res.set('Retry-After', String(secs));
+		throw new HitRateLimitException(`Too many requests, retry after ${secs}.`);
+	}
```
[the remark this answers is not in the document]
kukulaka: Please see https://github.com/animir/node-rate-limiter-flexible/wiki/API-methods#ratelimiterconsumekey-points--1-options--

rijkvanzanten (approved):

PR #151, opened 2020-08-18
Feature redis cache

kukulaka:
merging from rate limiter


api/src/controllers/activity.ts
```diff
@@ -1,24 +1,61 @@
...
+				const cacheService = new CacheService(TTLnum, dTTL);
+				cacheService.setCache(
+					key,
+					JSON.stringify({
+						data: records || null,
+						meta,
+					})
+				);
+			}
+		}
```
rijkvanzanten, line 58: Do we have to do this stuff in every controller? Can't the `cacheMiddleware` handle this?

rijkvanzanten: Alternatively, can we change this to a before/after middleware structure?

```
...

sanitizeQuery,

checkCacheMiddleware,

asyncHandler(async ...controller...),

saveToCacheMiddleware

...
```

api/src/controllers/activity.ts
```diff
@@ -30,10 +67,35 @@ router.get(
 	'/:pk',
 	useCollection('directus_activity'),
 	sanitizeQuery,
+	cacheMiddleware,
 	asyncHandler(async (req, res) => {
+		const key = req.url;
+		const TTL = req.query.TTL;
+		const TTLnum = Number(TTL);
+		const dTTL = Number(req.query.dTTL);
```
rijkvanzanten, line 75: We'll have to think through how much flexibility we want to give through the query param vs a single config. By allowing somebody to set the TTL to 0, that could potentially open up a vector for DDoSSing (requesting something super heavy, and force it to skip the cache).

Also, ideally, the cache is invalidated on change (create/update/delete) of the collection, so being able to skip the cache might not event be relevant in that case :)

api/src/services/node-cache.ts
```diff
@@ -0,0 +1,53 @@
+/** cache service.
```
rijkvanzanten, line 1: Why is this service called `node-cache`? Little confusing seeing that the library is also called `node-cache`. Lets just call it `/services/cache.ts` / `CacheService`

api/src/services/node-cache.ts
```diff
@@ -0,0 +1,53 @@
...
+		if (value) {
+			return Promise.resolve(value);
+		}
+	}
+
+	async setCache(key: string, data: {}) {
+		this.apiCache.set(key, data);
+	}
+
+	// this flushes all data. important incase cache gets too full.
```
rijkvanzanten, line 49: When is it _too full_ / how / where is it flushed currently?

api/src/cli/utils/create-env/index.ts
```diff
@@ -25,6 +25,23 @@ const defaults = {
 		STORAGE_LOCAL_DRIVER: 'local',
 		STORAGE_LOCAL_ROOT: './uploads',
 	},
+	redisServer: {
+		REDIS_HOST: '127.0.0.1',
+		REDIS_PORT: '6379',
+		REDIS_PASSWORD: null,
+	},
+	rateLimits: {
```
rijkvanzanten, line 33: For the next PR, it would be nice if the branch is created from `main` instead of whatever the last thing was you were working on. Having the rate limit stuff intertwined makes it a bit harder to review 😁

api/src/controllers/activity.ts
rijkvanzanten: Alternatively, we could pull from / save to cache _in_ the `ItemsService`, but I don't know what the implications are of that elsewhere 🤔

rijkvanzanten:
Let's drop support for the `ttl` / `skip-cache` etc query param in favor of invalidating the cache on create/update/delete of a given collection.

Maybe we _have to_ do the caching in the ItemService, cause even with cache enabled, the API should still ensure that you have permissions to read the requested data. The permissions for the given user could have changed between requests. 

We should save the cache not just based on requested asset, but also based on the user and exact query params used. Otherwise we can end up in a situation where an admin requests a certain set of data (+ populates the cache), and a regular user gets the admin data from cache.

benhaynes:
Is this smart enough to invalidate when you `update` data in the database directly? We'll want to make sure we can invalidate as needed.... I like the idea of query params, but defer to Rijk.

rijkvanzanten:
In that case, the regular configured TTL will invalidate the cache within the configured time. If you're expecting to do a lot of direct-to-DB updates, you'd want to set the TTL to a low value, or disable caching completely.

benhaynes:
And the App will have a way to bypass all of this?

rijkvanzanten:
> And the App will have a way to bypass all of this?

By invalidating the cache on changes (create/update/delete), the app will never read outdated data and therefore doesn't have to rely on any special treatment

benhaynes:
But again, if you have a 5m cache, and have people/services updating data directly through SQL (in the DB) then the App would be outdated. Right?

rijkvanzanten:
In that case, the regular configured TTL will invalidate the cache within the configured time. If you're expecting to do a lot of direct-to-DB updates, you'd want to set the TTL to a low value, or disable caching completely.

🔁

benhaynes:
OK! I guess we can always add a query param option in the future as needed.

api/src/cli/utils/create-env/index.ts
kukulaka: eeek tdid mean to branch off main

rijkvanzanten (approved):

PR #153, opened 2020-08-19
Add display template interface

Nitwel:
I left the wrapper out for now and used the component directly but if you @rijkvanzanten prefer the use of an wrapper, I can definitely add that too.


app/src/interfaces/display-template/index.ts
```diff
@@ -0,0 +1,42 @@
...
+		{
+			field: 'disabled',
+			type: 'boolean',
+			name: i18n.t('disabled'),
+			meta: {
+				width: 'half',
+				interface: 'toggle',
+				default_value: false,
+			},
+		},
```
rijkvanzanten, line 30: Value / disabled are props of every interface and aren't interface options

app/src/interfaces/display-template/index.ts
```diff
@@ -0,0 +1,42 @@
...
+		{
+			field: 'collection',
+			name: i18n.t('collection'),
+			type: 'string',
+			meta: {
+				width: 'half',
+				interface: 'text-input',
+				default_value: '',
```
rijkvanzanten, line 38: Default value goes under `schema`, not `meta`

app/src/interfaces/display-template/display-template.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+	<v-field-template />
```
rijkvanzanten, line 2: Don't forget to pass `collection` to the v-field-template component 🙂

rijkvanzanten:
> prefer the use of an wrapper

No need for a wrapper element 👍

app/src/interfaces/display-template/display-template.vue
Nitwel: No need when this wrapper is not used. 😄

Nitwel:
This should now be ready to be merged.

PR #204, opened 2020-09-04
Current route API preview

Kematia:
Hey! Did some work on the Preview API functionality, I'd love to get some feedback from the team...

The position of the modal activator is a placeholder, we could place it in the upcoming Export Data section in the sidebar, as @benhaynes suggested!

Couple screenshots...
image
image
image

### DIFF
--- a/app/src/views/private/components/api-preview/api-preview.vue
+++ b/app/src/views/private/components/api-preview/api-preview.vue
@@ -1,0 +1,144 @@
+<template>
+	<v-modal title="Preview API Request" :subtitle="path">
+		<template #activator="{ on }">
+			<v-sheet style="text-align: center">
+				<v-button small outlined @click="on">Preview API Response</v-button>
+			</v-sheet>
+		</template>
+
+		<v-form v-model="formModel" :fields="fields" primaryKey="preview" @input="previewApiResponse()" />
+
+		<interface-code
+			v-bind="{ readOnly: true, language: 'json' }"
+			class="response-editor"
+			:value="JSON.stringify(responseModel, null, 5)"
+		/>
+
+		<template #footer="{ close }">
+			<v-button @click="close">Close</v-button>
+		</template>
+	</v-modal>
+</template>
+
+<style lang="scss" scoped>
+.response-editor {
+	padding-top: 20px;
+}
+</style>
+
+<script lang="ts">
+import { AxiosRequestConfig } from 'axios';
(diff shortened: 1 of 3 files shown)


rijkvanzanten:
I have a small concern about the 80/20-ness of this feature within the context of the sidebar on the browse page in regular use, but I think this is a superb addition to have under settings for administrator users 🤔 Thoughts on where this should live @benhaynes ?

benhaynes:
First off, thank you @Kematia — this is awesome! 😍 

I agree with @rijkvanzanten — this is certainly useful, but a very power-user sort of feature. No matter where it goes, I think it needs to be admin-only.

If we all decide this is a bit too "heavy" for the Browse/Detail page, then I can certainly see it as a section of the Settings... perhaps as part of the dynamic API Reference Docs (as Rijk mentioned).

But I kinda like the idea of having this on Browse/Detail — perhaps as an admin-only link/button (eg: `Explore API Data`) within the upcoming "Data Export" sidebar component? Here's a quick mockup I made (second button opens up this modal):

Item Browse Sidebar Filter

Kematia:
> First off, thank you @Kematia — this is awesome! 😍
> 
> I agree with @rijkvanzanten — this is certainly useful, but a very power-user sort of feature. No matter where it goes, I think it needs to be admin-only.
> 
> If we all decide this is a bit too "heavy" for the Browse/Detail page, then I can certainly see it as a section of the Settings... perhaps as part of the dynamic API Reference Docs (as Rijk mentioned).
> 
> But I kinda like the idea of having this on Browse/Detail — perhaps as an admin-only link/button (eg: `Explore API Data`) within the upcoming "Data Export" sidebar component? Here's a quick mockup I made (second button opens up this modal):
> 
> Item Browse Sidebar Filter
This is pretty much how I envisioned it, either admin-only or user-enabled

Kematia:
As of now the api route to preview is discenred from the route, which I think is something I'd liked to keep for ease and speed of use, but it would be easy to expand the interface to accept a user-specified route/collection to preview.

This would unlock a more "explorer" like browsing and, something I'd really love, the ability to preview the collection you're building in the settings/data_model view.

rijkvanzanten:
Thanks for the nice start @Kematia 🙂 We've moved the API testing IO into a dedicated section in the new docs module, so this PR has been included in and superseded by https://github.com/directus/next/pull/397 ❤️

PR #212, opened 2020-09-07
add v-model to tab-items

Kematia:
I need to add the v-model to v-tabs-items as well, is this correct? It does not work without it.

image


rijkvanzanten:
Yep you're right! Good catch

PR #220, opened 2020-09-08
Cleaned up all interfaces and displays.

Nitwel:


rijkvanzanten:
Can you resolve the merge conflict @Nitwel?

Nitwel:
Sure!

app/src/interfaces/code/code.vue
```diff
@@ -268,7 +274,12 @@ export default defineComponent({
 		};
 
 		function fillTemplate() {
-			emit('input', props.template);
+			try {
+				const parse = JSON.parse(props.template);
+				emit('input', parse || props.template);
+			} catch {
+				// We won't stage invalid JSON
+			}
```
rijkvanzanten, line 282: The template can be non-JSON types as well.

app/src/interfaces/code/code.vue
```diff
@@ -291,6 +302,12 @@ export default defineComponent({
 	text-align: right;
 }
 
+::v-deep .CodeMirror {
+	.CodeMirror-placeholder {
+		color: var(--foreground-subdued);
+	}
+}
+
```
rijkvanzanten, line 310: I believe there's a separate code mirror styles file

app/src/interfaces/code/index.ts
```diff
@@ -55,4 +57,5 @@ export default defineInterface(({ i18n }) => ({
 			},
 		},
 	],
+	recommendedDisplays: ['raw'],
```
rijkvanzanten, line 60: Raw display is forced recommended for every display. Can be omitted here

app/src/interfaces/color/index.ts
```diff
@@ -3,19 +3,22 @@ import InterfaceColor from './color.vue';
 
 export default defineInterface(({ i18n }) => ({
 	id: 'color',
-	name: i18n.t('color'),
+	name: i18n.t('interfaces.color.color'),
+	description: i18n.t('interfaces.color.description'),
 	icon: 'palette',
 	component: InterfaceColor,
 	types: ['string'],
+	recommendedDisplays: ['color-dot'],
```
rijkvanzanten, line 11: I don't know how I feel about having color-dot support both a "dynamic value" (eg when value = `published`, show blue) _and_ have it support a hex value to render 🤔 

I know the output is virtually identical (a colored dot), but the value it (re)presents is very different.. Thoughts @benhaynes?

app/src/interfaces/display-template/index.ts
```diff
@@ -3,10 +3,27 @@ import InterfaceDisplayTemplate from './display-template.vue';
...
+				interface: 'collections',
+				options: {
+					includeSystem: true,
+				},
+			},
+			schema: {
+				default_value: null,
+			},
+		},
+	],
```
rijkvanzanten, line 28: The display template interface doesn't have a `collection` option

app/src/interfaces/divider/divider.vue
```diff
@@ -7,6 +7,7 @@
 			'--v-divider-label-color': color,
 		}"
 		large
+		:inlineTitle="inlineTitle"
```
rijkvanzanten, line 10: Use html-attr style:

```suggestion
		:inline-title="inlineTitle"
```

app/src/interfaces/divider/index.ts
```diff
@@ -3,12 +3,14 @@ import InterfaceDivider from './divider.vue';
...
 	id: 'divider',
-	name: i18n.t('divider'),
+	name: i18n.t('interfaces.divider.divider'),
+	description: i18n.t('interfaces.divider.description'),
 	icon: 'remove',
 	component: InterfaceDivider,
 	hideLabel: true,
 	hideLoader: true,
 	types: ['alias'],
+	recommendedDisplays: ['raw'],
```
rijkvanzanten, line 13: Raw is always recommended

rijkvanzanten: Also, displays aren't visible for presentation fields I don't think 🤔 (shouldn't be at least)

app/src/interfaces/markdown/index.ts
```diff
@@ -15,16 +16,26 @@ export default defineInterface(({ i18n }) => ({
...
+				options: {
+					label: i18n.t('interfaces.markdown.tabbed_label'),
+				},
+			},
+			schema: {
+				default_value: false,
 			},
 		},
 	],
+	recommendedDisplays: ['raw'],
```
rijkvanzanten, line 40: raw is always recommended

app/src/interfaces/notice/index.ts
```diff
@@ -3,12 +3,14 @@ import InterfaceNotice from './notice.vue';
...
-	name: i18n.t('notice'),
-	icon: 'remove',
+	name: i18n.t('interfaces.notice.notice'),
+	description: i18n.t('interfaces.notice.description'),
+	icon: 'info',
 	component: InterfaceNotice,
 	hideLabel: true,
 	hideLoader: true,
 	types: ['alias'],
+	recommendedDisplays: ['raw'],
```
rijkvanzanten, line 13: raw is always recommeded

app/src/interfaces/radio-buttons/index.ts
```diff
@@ -49,9 +53,6 @@ export default defineInterface(({ i18n }) => ({
 				width: 'half',
 				interface: 'icon',
 			},
-			schema: {
-				default_value: 'radio_button_checked',
-			},
```
rijkvanzanten, line 54: Lets leave this in so it's immediately clear what the default value is when setting up the options

app/src/interfaces/radio-buttons/index.ts
```diff
@@ -61,9 +62,6 @@ export default defineInterface(({ i18n }) => ({
 				width: 'half',
 				interface: 'icon',
 			},
-			schema: {
-				default_value: 'radio_button_unchecked',
-			},
```
rijkvanzanten, line 66: Lets leave this in so it's immediately clear what the default value is when setting up the options

app/src/interfaces/repeater/index.ts
```diff
@@ -4,9 +4,11 @@ import RepeaterOptions from './options.vue';
...
 export default defineInterface(({ i18n }) => ({
 	id: 'repeater',
-	name: i18n.t('repeater'),
+	name: i18n.t('interfaces.repeater.repeater'),
+	description: i18n.t('interfaces.repeater.description'),
 	icon: 'replay',
 	component: InterfaceRepeater,
 	types: ['json'],
 	options: RepeaterOptions,
+	recommendedDisplays: ['related-values'],
```
rijkvanzanten, line 13: Does the repeater (non-relational JSON) interface work with the related values display? I don't think it does

app/src/interfaces/repeater/options.vue
```diff
@@ -1,10 +1,29 @@
...
+			<div class="grid-element">
+				<p class="type-label">{{ $t('interfaces.repeater.max_amount') }}</p>
+				<v-input
+					class="input"
+					type="number"
+					v-model="limit"
+					:min="1"
+					:placeholder="$t('interfaces.repeater.max_amount_placeholder')"
+				/>
+			</div>
```
rijkvanzanten, line 17: We should discuss if and how we want to tackle this in the API / validation / permissions / etc. Seeing this is just a JSON blob, I don't think the API can / should do any validation for this. (@benhaynes)

app/src/interfaces/repeater/repeater.vue
```diff
@@ -42,7 +42,7 @@ export default defineComponent({
 			type: String,
 			required: true,
 		},
-		createItemText: {
+		placeholder: {
```
rijkvanzanten, line 45: A placeholder is not the same as the text that goes in the create-new-item button 🤔 

The placeholder would be the text that shows in the rows when they're empty IMO

app/src/interfaces/repeater/options.vue
```diff
@@ -127,17 +158,25 @@ export default defineComponent({
...
+	display: grid;
+	grid-template-columns: [start] minmax(0, 1fr) [half] minmax(0, 1fr) [full];
+	gap: var(--form-vertical-gap) var(--form-horizontal-gap);
+
+	&-element {
+		&.full {
+			grid-column: start/full;
+		}
+	}
 }
```
rijkvanzanten, line 177: Would be nice if we could make this re-usable somehow 🤔 

Maybe worth making this a mixin, seeing there's multiple places where we want to/have to mimic the form without using `v-form`

app/src/interfaces/slider/index.ts
```diff
@@ -3,37 +3,39 @@ import { defineInterface } from '@/interfaces/define';
 
 export default defineInterface(({ i18n }) => ({
 	id: 'slider',
-	name: i18n.t('slider'),
+	name: i18n.t('interfaces.slider.slider'),
+	description: i18n.t('interfaces.slider.description'),
 	icon: 'linear_scale',
 	component: InterfaceSlider,
 	types: ['integer', 'decimal', 'float', 'bigInteger'],
+	recommendedDisplays: ['rating'],
```
rijkvanzanten, line 11: Not sure if rating is a good match for the `slider` interface 🤔 @benhaynes thoughts?

app/src/interfaces/slug/index.ts
```diff
@@ -3,9 +3,41 @@ import InterfaceSlug from './slug.vue';
...
 export default defineInterface(({ i18n }) => ({
 	id: 'slug',
-	name: i18n.t('slug'),
+	name: i18n.t('interfaces.slug.slug'),
+	description: i18n.t('interfaces.slug.description'),
 	icon: 'link',
 	component: InterfaceSlug,
 	types: ['string'],
-	options: [],
+	recommendedDisplays: ['formatted-value'],
```
rijkvanzanten, line 11: Formatted value is recommended for every interface

app/src/interfaces/slug/index.ts
```diff
@@ -3,9 +3,41 @@ import InterfaceSlug from './slug.vue';
...
+			field: 'placeholder',
+			name: i18n.t('placeholder'),
+			meta: {
+				width: 'half',
+				interface: 'text-input',
+				options: {
+					placeholder: i18n.t('enter_a_placeholder'),
+				},
+			},
+		},
```
rijkvanzanten, line 23: Lets make this one full width

app/src/interfaces/slug/index.ts
```diff
@@ -3,9 +3,41 @@ import InterfaceSlug from './slug.vue';
...
+		{
```
rijkvanzanten, line 24: That way it's 

```
[ Placeholder                                   ]
[ Icon (Left)        ]   [ Icon (Right)         ]
```

app/src/interfaces/tags/tags.vue
```diff
@@ -52,7 +52,7 @@ export default defineComponent({
 		},
 		value: {
 			type: Array as PropType<string[]>,
-			default: null,
+			default: () => [],
```
rijkvanzanten, line 55: We should expect the default value to be `null`. That's what it would be coming from the API as well

app/src/interfaces/text-input/index.ts
```diff
@@ -3,10 +3,12 @@ import { defineInterface } from '@/interfaces/define';
 
 export default defineInterface(({ i18n }) => ({
 	id: 'text-input',
-	name: i18n.t('text_input'),
+	name: i18n.t('interfaces.text-input.text-input'),
+	description: i18n.t('interfaces.text-input.description'),
 	icon: 'text_fields',
 	component: InterfaceTextInput,
 	types: ['string', 'uuid'],
+	recommendedDisplays: ['raw'],
```
rijkvanzanten, line 11: raw is always recommended

app/src/interfaces/text-input/index.ts
```diff
@@ -46,10 +52,41 @@ export default defineInterface(({ i18n }) => ({
...
+				options: {
+					label: i18n.t('interfaces.text-input.trim_label'),
+				},
+			},
+			schema: {
+				default_value: false,
+			},
+		},
+		{
+			field: 'mask',
```
rijkvanzanten, line 77: I believe the prop is called `masked` in the text-input interface

rijkvanzanten, line 77:
```suggestion
			field: 'masked',
```

app/src/interfaces/textarea/index.ts
```diff
@@ -3,35 +3,57 @@ import { defineInterface } from '@/interfaces/define';
 
 export default defineInterface(({ i18n }) => ({
 	id: 'textarea',
-	name: i18n.t('textarea'),
+	name: i18n.t('interfaces.textarea.textarea'),
+	description: i18n.t('interfaces.textarea.description'),
 	icon: 'text_fields',
 	component: InterfaceTextarea,
 	types: ['text'],
+	recommendedDisplays: ['raw'],
```
rijkvanzanten, line 11: raw is always recommended

app/src/interfaces/user/user.vue
```diff
@@ -102,7 +102,7 @@ export default defineComponent({
 	components: { ModalDetail, ModalBrowse },
 	props: {
 		value: {
-			type: [Number, Object],
+			type: [String, Object],
```
rijkvanzanten, line 105: The user interface will never get an Object for it's value

app/src/interfaces/wysiwyg/index.ts
```diff
@@ -3,14 +3,16 @@ import { defineInterface } from '@/interfaces/define';
 
 export default defineInterface(({ i18n }) => ({
 	id: 'wysiwyg',
-	name: i18n.t('wysiwyg'),
+	name: i18n.t('interfaces.wysiwyg.wysiwyg'),
+	description: i18n.t('interfaces.wysiwyg.description'),
 	icon: 'format_quote',
 	component: InterfaceWYSIWYG,
 	types: ['text'],
+	recommendedDisplays: ['formatted-value'],
```
rijkvanzanten, line 11: formatted value is always recommended

app/src/interfaces/wysiwyg/index.ts
```diff
@@ -36,6 +38,7 @@ export default defineInterface(({ i18n }) => ({
 				width: 'half',
 				interface: 'dropdown-multiselect',
 				options: {
+					closeOnContentClick: false,
```
rijkvanzanten, line 41: This shouldn't be an interface option. It should be disabled by default in the dropdown multiselect interface

app/src/modules/settings/routes/data-model/field-detail/components/display.vue
```diff
@@ -71,15 +72,23 @@ export default defineComponent({
...
-				return displayItems;
+				recommendedItems.push({ divider: true });
 			}
+
+			const displayList = displayItems.filter((item) => recommended.includes(item.value as string) === false);
+			if (displayList !== undefined) {
+				recommendedItems.push(...displayList.filter((i) => i));
+			}
+
+			return recommendedItems;
```
rijkvanzanten, line 91: What do these changes do?

Doesn't make it easier to read imo

app/src/modules/settings/routes/data-model/field-detail/components/interface.vue
```diff
@@ -85,15 +85,23 @@ export default defineComponent({
...
+
 			if (interfaceItems.length >= 5 && recommended.length > 0) {
-				return [
-					...recommended.map((key) => interfaceItems.find((item) => item.value === key)),
-					{ divider: true },
-					...interfaceItems.filter((item) => recommended.includes(item.value as string) === false),
-				].filter((i) => i);
-			} else {
-				return interfaceItems;
+				recommendedItems.push({ divider: true });
```
rijkvanzanten, line 96: Not sure if this is any easier to read that what was there before

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
```diff
@@ -40,11 +40,45 @@
...
+					class="monospace"
+					v-model="default_value"
+					:placeholder="$t('add_a_default_value')"
+				/>
+				<v-textarea
+					v-else-if="default_type === 'text' || default_type === 'json'"
+					class="monospace"
+					v-model="default_value"
+					:placeholder="$t('add_a_default_value')"
+				/>
```
rijkvanzanten, line 56: textarea is bigger than other interfaces, we'll have to double check how that jumps on the page when switching between types. Maybe we'll have to make the default value full width 
at all times (@benhaynes )

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
```diff
@@ -40,11 +40,45 @@
...
-					v-model="fieldData.schema.default_value"
+					v-model="default_value"
+					:placeholder="$t('add_a_default_value')"
+				/>
+				<v-input
+					v-else-if="default_type === 'datetime'"
+					class="monospace"
+					v-model="default_value"
+					:placeholder="$t('add_a_default_value')"
+				/>
```
rijkvanzanten, line 69: Should we use the datetime interface here?

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
```diff
@@ -201,13 +235,39 @@ export default defineComponent({
 			return i18n.t('choose_a_type');
 		});
 
-		return { fieldData: state.fieldData, typesWithLabels, setType, typeDisabled, typePlaceholder };
+		const default_value = computed({
```
rijkvanzanten, line 238: camelCase

```suggestion
		const defaultValue = computed({
```

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
```diff
@@ -201,13 +235,39 @@ export default defineComponent({
...
+			get() {
+				return state.fieldData.schema.default_value;
+			},
+			set(newVal: any) {
+				state.fieldData.schema.default_value = newVal;
+			},
+		});
+
+		const default_type = computed(() => {
```
rijkvanzanten, line 247: What is this? What does it do?

rijkvanzanten: Oh, it's the scoped type you're using for the default value if statement above. I would change that to be inline honestly:

Before:

```
v-else-if="default_type === 'number'"
```

After:

```
v-else-if="['bigInteger', 'integer', 'float', 'decimal'].includes(state.fieldData.type)"
```

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
```diff
@@ -201,13 +235,39 @@ export default defineComponent({
...
+		});
+
+		return {
+			fieldData: state.fieldData,
+			typesWithLabels,
+			setType,
+			typeDisabled,
+			typePlaceholder,
+			default_value,
+			default_type,
```
rijkvanzanten, line 265: camelCase

app/src/modules/settings/routes/data-model/field-detail/store.ts
```diff
@@ -291,6 +291,25 @@ function initLocalStore(
...
+		watch(
+			() => state.fieldData.type,
+			() => {
+				switch (state.fieldData.type) {
+					case 'json':
+						state.fieldData.meta.special = 'json';
+						break;
+					case 'boolean':
+						state.fieldData.meta.special = 'boolean';
+						state.fieldData.schema.is_nullable = false;
```
rijkvanzanten, line 305: Lets also enforce the default value to be set if you explicitly set `is_nullable` to false

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
```diff
@@ -201,13 +235,39 @@ export default defineComponent({
...
+		};
 
 		function setType(value: typeof types[number]) {
 			if (value === 'uuid') {
 				state.fieldData.meta.special = 'uuid';
```
rijkvanzanten, line 270: Lets remove this from here as well in favor of doing all this in the store.ts file

rijkvanzanten (changes requested):
I think we should make placeholders full width in all options. @benhaynes thoughts?

benhaynes:
Yes, certain options should always be full-width... and `placeholder` is one of them.

app/src/interfaces/color/index.ts
benhaynes: Hmm, interesting. 

* **Color Dot** based on hex value (eg: color interface)
* **Color Dot** based on dynamic value (eg: category/status)
* **Label** based on dynamic value (eg: category/status)

I would say that we could either combine Color Dot and Label (making a "Show as Dot" option like in v8, or we need to rename these to be clearer (eg: `Label Dot` and `Color`).

rijkvanzanten: > making a "Show as Dot" option

My vote goes to this ☝️

app/src/interfaces/repeater/options.vue
benhaynes: Very useful/powerful feature(s) to have... and it makes sense to set them up in the interface options. That said, I agree that these should (ideally) be enforced by the API too. We've been trying to think of a way to make all of these validation rules global per field... that seems like the cleanest way forward.

The only exception is if the interface validation is really APP ONLY... can't think of any examples of that though.

app/src/interfaces/slider/index.ts
benhaynes: Hmm, I think it's a good option, but not recommended.

app/src/interfaces/repeater/options.vue
rijkvanzanten: Lets remove the "max_amount" option from the repeater options for the time being, until we figure out how to tackle this "for real" on the field level (w/ API support)

app/src/interfaces/color/index.ts
benhaynes: Agree — v8 for the win! We don't have conditional fields, but it SHOULD be clear that "Foreground Color" won't be used in "Dot Mode". I think this new structure makes sense given that we're converting this to be called `Label` now too.

app/src/interfaces/display-template/index.ts
Nitwel: collectionField prop of the interface should be it. But what exactly means `collectionField` / what does the prop want to get?

app/src/interfaces/divider/index.ts
Nitwel: As far as I remember they are.

app/src/interfaces/repeater/index.ts
Nitwel: Ah yes that was my bad. I thought relational would also support json format! I guess we currently have no proper way of displaying json?

app/src/interfaces/repeater/repeater.vue
Nitwel: Well, it is a placeholder. It is the text of the placeholder for a new Item. And for `createItemText`, it is not very obvious what it does. (You could understand it as a option to create text for an item)
So placeholder is the way to go IMO.

app/src/interfaces/repeater/options.vue
Nitwel: 💯 I just don't know how to that cause v-form is very complex aready.

app/src/modules/settings/routes/data-model/field-detail/components/interface.vue
Nitwel: I changed it so the order always stays the same only the divider does not get rendered when below 5 elements.
The reason for this is that it's disorientating when the recomended interface comes last.

app/src/interfaces/display-template/index.ts
rijkvanzanten: It requires a field name (string) of which the value is the collection

app/src/interfaces/repeater/index.ts
rijkvanzanten: Not really 🤔 Feel free to start a discussion if you have any thoughts about what a good display for JSON would be

app/src/interfaces/repeater/repeater.vue
rijkvanzanten: > It is the text of the placeholder for a new Item.

.. except it's not 😄 

CleanShot 2020-09-09 at 13 09 38

(1) = placeholder
(2) = create item text

app/src/interfaces/display-template/index.ts
Nitwel: Shouldn't we rename the prop to collectionName or to collection then cause collectionField is confusable with a field of the collection. (Thats what confused me)

app/src/interfaces/repeater/repeater.vue
Nitwel: I still think it's quite confusing. 🤷‍♂️

app/src/interfaces/display-template/index.ts
rijkvanzanten: > Shouldn't we rename the prop to collectionName or to collection

No, cause the value isn't the name of a collection, the value is the name of another field on the page that in turn holds the collection name. So:

```
collection_name
[ text-input                               ]


template
[ display-template                         ]


Options:
template.collectionField = collection_name
```

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
Nitwel: Sounds good. How is the interface vue-component called internally?

app/src/interfaces/repeater/repeater.vue
rijkvanzanten: Sure, but renaming it `placeholder` is not the solution, as `placeholder` is something else in this context. We can workshop names though, would `ctaText` be better? `ctaText`, `createLabel`, `addLabel`, ?

app/src/modules/settings/routes/data-model/field-detail/store.ts
Nitwel: Any ideas how to do this. Is there a required option? I currently only can think of letting the user not go to the next section.

app/src/interfaces/repeater/repeater.vue
Nitwel: I really like addLabel but in the end it's up to you!

app/src/interfaces/display-template/index.ts
Nitwel: >It requires a field name (string) of which the value is the collection
Read that wrong.

app/src/modules/settings/routes/data-model/field-detail/components/schema.vue
Nitwel: I think you overlooked line 43. 😉 
grafik

app/src/modules/settings/routes/data-model/field-detail/store.ts
rijkvanzanten: We could just set 

```ts
state.fieldData.schema.default_value = false;
```

whenever you change to a boolean type! That way we at least know for sure that a default value is set, and therefore `nullable` is met

app/src/interfaces/code/code.vue
```diff
@@ -268,7 +274,11 @@ export default defineComponent({
 		};
 
 		function fillTemplate() {
-			emit('input', props.template);
+			try {
+				emit('input', JSON.parse(props.template));
+			} catch {
+				emit('input', props.template);
+			}
```
rijkvanzanten, line 281: Don't hate me for this, but:

JSON is also valid JS, but we don't want to send a JSON object to the API when JavaScript is used as the language. We should check for props.type and only try JSON.parse when props.type == json

Nitwel: 🤦 Yea right!

rijkvanzanten (changes requested):

rijkvanzanten (approved):

rijkvanzanten (approved):

Issue #232, opened 2020-09-09
Datetime interface enables save without changes on existing items


benhaynes:
Edit page with DateTime interface has enabled save button changes (existing items only)

 Requires reworking staging/conversion logic (1h)

Nitwel:
Gota fix this real quick!

PR #240, opened 2020-09-09
fix datetime emiting without changes

Nitwel:
Fixes https://github.com/directus/next/issues/232

Had to check if the dates match cause the iso format was formatted as a different string which still was correct iso.


rijkvanzanten (approved):

benhaynes:
Yaassss!!

PR #273, opened 2020-09-10
Hide too many checkboxes

Nitwel:
This image shows how I implemented this but I'm up for a discussion on how we want to toggle this!
grafik

### DIFF
--- a/app/src/components/v-detail/readme.md
+++ b/app/src/components/v-detail/readme.md
@@ -10,9 +10,10 @@ Allows for collapsable content
 
 ## Props
 
-| Prop     | Description         | Default |
-|----------|---------------------|---------|
-| `active` | Used with `v-model` | `false` |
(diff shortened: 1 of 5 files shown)


rijkvanzanten:
Lets make the toggle like this:

CleanShot 2020-09-10 at 09 54 22

rijkvanzanten:
@benhaynes Should we make this an option? I can imagine you might want to force it to show all of them at all times if you have 8–12 items (it cuts off at 7 now)

Nitwel:
That looks better!

benhaynes:
Yeah, I think it should be an option — could be a toggle or an INT of when to cut-off.

Nitwel:
This is how it looks now:
grafik

And with custom options:
grafik

Nitwel:
This should be ready for review now!

rijkvanzanten (approved):

PR #277, opened 2020-09-10
Fix md interface

Nitwel:


Nitwel:
Now looks like this.
grafik

app/src/interfaces/markdown/markdown.vue
```diff
@@ -62,7 +62,11 @@ export default defineComponent({
 		const showEdit = computed(() => !props.tabbed || currentTab.value[0] === 0);
 		const showPreview = computed(() => !props.tabbed || currentTab.value[0] !== 0);
 
-		onMounted(updateHasScrollbar);
+		onMounted(() => {
+			setTimeout(() => {
+				updateHasScrollbar();
+			}, 100);
```
rijkvanzanten, line 68: Why wait exactly 100ms? Feels like you're waiting for some other process to finish

Nitwel: I don't really know what im wating here for.
I want to get the text with the scrollbar rendered but apparently this is not the case after mounted so I have to wait longer.

rijkvanzanten: Ok, so lets figure out what the root cause is. This is a bit of a nasty hack to get around it 🙂

Nitwel: Mounted does call atfer all child components have been mounted right?

rijkvanzanten: I'm not sure, but I believe it does, yes

rijkvanzanten: @Nitwel 

> I don't really know what im wating here for.

Lets make sure we know what we're waiting for, and come up with a solution to that problem

rijkvanzanten:
Instead of having an expand toggle, lets make this work the same as the WYSIWYG, where it auto-expands whenever the content starts overflowing

Nitwel:
Yep, this branch can be closed I think cause it is a feature required in many situations but implemented for single use here.
So either we find a general way of doing this or have it auto expand when writing.

PR #281, opened 2020-09-11
Backup databases

kukulaka:


kukulaka:
Having trouble in windows streaming the generated file. (fs not finding file) I think it might be a permissions thing

kukulaka:
Have confirmed as a permissions issue. When moved to a different location or used path then worked. @rijkvanzanten any ideas on this? When moved export fold to within src the export worked, but when above src (or elsewhere) does not. (at least in windows) .

We have to export file initially to a location on the server as all the export tools use tools installed on the server.

kukulaka:
Actually could just store the db in local with everything else? Then would get rid of this issue @rijkvanzanten

rijkvanzanten:
Yeah, that works. What we can do (similar to extensions), is have a single configuration option that says where to store the backups, and use that for the IO here

PR #287, opened 2020-09-11
suggest fields on m2o

Nitwel:
Adds the v-field-template to the m2o interface.
I also implemented a depth prop for that component.

grafik


app/src/components/v-field-template/field-list-item.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<v-list-item
-		v-if="field.children === undefined"
+		v-if="field.children === undefined || depth == 0"
```
rijkvanzanten, line 3:
```suggestion
		v-if="field.children === undefined || depth === 0"
```

app/src/components/v-field-template/field-list-item.vue
```diff
@@ -32,6 +33,10 @@ export default defineComponent({
 			type: String,
 			default: null,
 		},
+		depth: {
+			type: Number,
+			default: 10,
```
rijkvanzanten, line 38: Lets make this something more sensical, like 2

rijkvanzanten:
```suggestion
			default: 2,
```

app/src/components/v-field-template/v-field-template.vue
```diff
@@ -49,6 +49,10 @@ export default defineComponent({
 			type: String,
 			required: true,
 		},
+		depth: {
+			type: Number,
+			default: 10,
```
rijkvanzanten, line 54:
```suggestion
			default: 2,
```

app/src/interfaces/many-to-one/options.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+	<div class="grid">
+		<v-notice type="warning" v-if="collection == null">
+			{{ $t('interfaces.one-to-many.no_collection') }}
+		</v-notice>
+		<div v-else class="full">
+			<p class="type-label">{{ $t('interfaces.many-to-one.display_template') }}</p>
+			<v-field-template :collection="collection" v-model="template" :depth="0"></v-field-template>
```
rijkvanzanten, line 8: The m2o support dot notation fields, so lets allow them to go a level deeper here

```suggestion
			<v-field-template :collection="collection" v-model="template" :depth="1" />
```

app/src/interfaces/many-to-one/options.vue
```diff
@@ -0,0 +1,60 @@
...
+			},
+			set(newTemplate: string) {
+				emit('input', {
+					...(props.value || {}),
+					template: newTemplate,
+				});
+			},
+		});
+		const collection = computed(() => {
+			if (props.fieldData.field == null || props.fieldData.meta?.collection == null) return null;
```
rijkvanzanten, line 42:
```suggestion
			if (props.fieldData.field === null || props.fieldData.meta?.collection === null) return null;
```

app/src/interfaces/many-to-one/options.vue
```diff
@@ -0,0 +1,60 @@
...
+			const relationData = relationsStore.getRelationsForField(
+				props.fieldData.meta.collection,
+				props.fieldData.field
+			)?.[0];
```
rijkvanzanten, line 46: It's not a given that the first relation is the one we're looking for. We have to explicitly match many_field and many_collection to ensure this is the correct one

rijkvanzanten (changes requested):

rijkvanzanten (approved):
LGTM!﻿

PR #294, opened 2020-09-14
fix focus border and custom formats

Nitwel:


app/src/interfaces/wysiwyg/wysiwyg.vue
```diff
@@ -1,5 +1,11 @@
 <template>
-	<Editor ref="editorElement" :init="editorOptions" v-model="_value" />
+	<Editor
+		ref="editorElement"
+		:init="editorOptions"
+		v-model="_value"
+		@onFocusIn="setFocus(true)"
+		@onFocusOut="setFocus(false)"
```
rijkvanzanten, line 7: This feels very trash, but I guess we gotta do what we gotta do here..

(maybe it's time to swap out "vue-tinymce" for the regular tinymce, so we don't have to hack around these extra "niceties") (that should also prevent that weird error on unmount)

rijkvanzanten (approved):

PR #295, opened 2020-09-14
Add missing chooseFromLib and loadFromUrl options in modal

Nitwel:
grafik

### DIFF
--- a/app/src/interfaces/image/image.vue
+++ b/app/src/interfaces/image/image.vue
@@ -45,53 +45,7 @@
 			/>
 			<file-lightbox v-model="lightboxActive" :id="image.id" />
 		</div>
-		<template v-else>
-			<v-upload @upload="setImage" />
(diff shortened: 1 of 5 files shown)


app/src/interfaces/image/image.vue
```diff
@@ -301,4 +402,17 @@ img {
 .disabled-placeholder {
 	height: var(--input-height-tall);
 }
+
+.options {
+	position: absolute;
+	top: 13px;
+	right: 13px;
```
rijkvanzanten, line 409: We're using multiples of 4 everywhere else. Lets make this 12px

rijkvanzanten:
I feel like we should add this to v-upload (w/ a prop to turn it on/off) instead of the image interface 🤔 

I think we want this on "new file" modal as well (in file library), and I can imagine we also want this when adding a new file through a/the multiple images interface

rijkvanzanten (approved):

PR #304, opened 2020-09-15
Added case and space-replace prop to tags interface

Nitwel:
grafik

### DIFF
--- a/app/src/interfaces/tags/tags.vue
+++ b/app/src/interfaces/tags/tags.vue
@@ -63,7 +63,7 @@ export default defineComponent({
 			type: String,
 			default: null,
 		},
-		case: {
+		capitalization: {
 			type: String as PropType<'uppercase' | 'lowercase' | 'auto-format'>,
 			default: null,
 		},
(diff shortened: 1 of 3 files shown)


rijkvanzanten:
@benhaynes Thoughts on the "replace space" option? I feel like we might want to change it to "db safe" so it matches a little closer to what we're doing elsewhere (force lowercase + underscores)

benhaynes:
I like the granularity of this... since it's not really about being DB safe, and more about having consistent tag formatting for your specific use-case. Although I would change the label to `Whitespace` and then have the action in the dropdown:

* `Leave as-in`
* `Replace with underscore`
* `Replace with hyphen`

Also, let's use `capitalization` instead of `case`.

Lastly, let's make sure OUR capitalization is consistent for all these options. For instance:

* `Force Alphabetical Order`
* `Allow other values`

app/src/lang/en-US/interfaces.json
```diff
@@ -152,8 +152,14 @@
 		"tags": {
 			"tags": "Tags",
 			"description": "Select or add tags",
-			"lowercase": "Lowercase",
-			"lowercase_label": "Force Lowercase",
+			"space_replace": "Whitespace",
```
rijkvanzanten, line 155: Can you keep the translations key inline with the actual value? 

`space_replace` shouldn't be rendered as `Whitespace`. `space_replace` = `Space Replace`, `whitespace` = `Whitespace`

app/src/interfaces/tags/tags.vue
```diff
@@ -160,9 +177,9 @@ export default defineComponent({
 		}
 
 		function removeTag(tag: string) {
-			selectedValsLocal.value = selectedValsLocal.value.filter(
-				(savedTag) => (props.lowercase ? savedTag.toLowerCase() : savedTag) !== tag
-			);
+			console.log(tag);
```
rijkvanzanten, line 180: Found one!

rijkvanzanten (approved):

rijkvanzanten:
@Nitwel Lets clear out that stray console log, and then we're good to merge 👍

app/src/interfaces/tags/tags.vue
Nitwel: Upsie!

Issue #309, opened 2020-09-15
MySQL datetime is being passed wrong value


shealavington:
I can't save any new items to my collection because of the following error, could be related to MySQL 5.7 only.

MySQL Ver 14.14 Distrib **5.7.31**, for Linux (x86_64) using  EditLine wrapper

```
{
  "errors": [
    {
      "message": "insert into `articles` (`content`, `date_created`, `slug`, `title`, `user_created`) values ('<p>z</p>', '\\\"2020-09-15T17:19:40.420Z\\\"', 'x', 'y', '6ef9c168-228b-4224-a1a5-fa4038dbcab0') 
      - ER_TRUNCATED_WRONG_VALUE: Incorrect datetime value: '\"2020-09-15T17:19:40.420Z\"' for column 'date_created' at row 1",
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"
      }
    }
  ]
}
```

shealavington:
That datetime issue I'm having, looks like it's attempting to save `'\\\"2020-09-15T17:19:40.420Z\\\"'` as the value, I would imagine it should be `'2020-09-15T17:19:40.420Z'`, no? :see_no_evil:

rijkvanzanten:
This was fixed in https://github.com/directus/next/commit/d87fc912f4fd095d9dfe2c33bee678578db2b919

PR #317, opened 2020-09-17
Regex JS Date to match timestamp format

shealavington:
https://github.com/directus/next/issues/309


api/src/services/payload.ts
```diff
@@ -175,9 +175,10 @@ export default class PayloadService {
 
 		if (['create', 'update'].includes(action)) {
 			processedPayload.forEach((record) => {
+				const jsDateRegex = /\"(\d{4}-\d{2}-\d{2})T(\d{2}:\d{2}:\d{2})(?:\.\d{3}Z)\"/
 				for (const [key, value] of Object.entries(record)) {
 					if (Array.isArray(value) || (typeof value === 'object' && value !== null)) {
-						record[key] = JSON.stringify(value);
+						record[key] = JSON.stringify(value).replace(jsDateRegex, '$1 $2')
```
rijkvanzanten, line 181: This is for nested objects in general, so we can't just replace every object with a date string.

Secondly, this date format is MySQL only, Postgres uses iso8601

rijkvanzanten:
Good attempt 🏅, but not the solution 🥉. Closing this for now

rijkvanzanten:
CleanShot 2020-09-18 at 12 56 27

This did the trick 😉

PR #380, opened 2020-09-21
force port to always be evaluated as number

kukulaka:
when config was generated, port was being parsed as string. MSSQL driver does not like this.


api/src/database/index.ts
```diff
@@ -18,9 +18,15 @@ for (let [key, value] of Object.entries(env)) {
...
 
-	connectionConfig[camelCase(key)] = value;
+	if (key === 'port') {
+		connectionConfig[camelCase(key)] = Number(value);
+	} else {
+		connectionConfig[camelCase(key)] = value;
+	}
 }
 
+console.log(connectionConfig);
```
rijkvanzanten: Sneaky console log!

api/src/database/index.ts
```diff
@@ -18,9 +18,15 @@ for (let [key, value] of Object.entries(env)) {
 
 	key = key.slice(3); // remove `DB_`
 
-	connectionConfig[camelCase(key)] = value;
+	if (key === 'port') {
+		connectionConfig[camelCase(key)] = Number(value);
+	} else {
+		connectionConfig[camelCase(key)] = value;
+	}
```
rijkvanzanten, line 25: I'm wondering if we should try this for any key in `/src/env.ts` instead. There's quite a couple "number" configurations (like rate limiter points) that would be convenient to convert ahead of time 🤔

rijkvanzanten (changes requested):

api/src/database/index.ts
kukulaka: @rijkvanzanten that sounds like a good idea to me 💯

api/src/database/index.ts
```diff
@@ -18,7 +19,11 @@ for (let [key, value] of Object.entries(env)) {
 
 	key = key.slice(3); // remove `DB_`
 
-	connectionConfig[camelCase(key)] = value;
+	if (isNumeric(value)) {
+		connectionConfig[camelCase(key)] = Number(value);
+	} else {
+		connectionConfig[camelCase(key)] = value;
+	}
```
rijkvanzanten, line 26: We should do this in `src/env.ts` instead 🙂  (across all env vars)

mscbpi: @rijkvanzanten  may you review https://github.com/directus/next/pull/623 ?

rijkvanzanten: I may! (no need to tag me btw, I see everything coming in/going out of directus/* 😉 )

rijkvanzanten:
Closed in favor of https://github.com/directus/next/pull/623

PR #405, opened 2020-09-22
Fix drag handles not working on Windows

### LINKED ISSUE #314: Collection columns: Drag handles not working
Handles don't work for dragging and re-ordering

image

nickrum:
Fixes #314

In theory this only hides the underlying issue. But investigating further seems to be a more complex task.

### DIFF
--- a/app/src/views/private/private-view.vue
+++ b/app/src/views/private/private-view.vue
@@ -157,17 +157,15 @@ export default defineComponent({
 			}
 
 			function onDragEnter(event: DragEvent) {
+				if (!event.dataTransfer) return;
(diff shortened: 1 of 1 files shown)


rijkvanzanten (approved):

rijkvanzanten:
Thanks @nickrum! I think we'll end up scoping the "global file drop" more tightly to just the browse of file library, which should also prevent this issue from popping up elsewhere. in the meantime, this is a perfectly reasonable fix IMO

nickrum:
Glad I could help! That sounds like a reasonable thing to do. In case the file library ever gets any drag handles, this fix might still be of some use.

PR #407, opened 2020-09-23
Adds container image pipeline

WoLfulus:
This should build images when new releases are pushed. 

It depends on Directus being published on npm first, so it will build only on release events.

Triggering on pushes would be nice in the future, but not implemented here.

This also adds exit signals processing into the `start` command, so containers can exit gracefully.

### DIFF
--- a/.github/actions/build-images/rootfs/directus/images/main/Dockerfile
+++ b/.github/actions/build-images/rootfs/directus/images/main/Dockerfile
@@ -26,14 +26,16 @@ RUN cat package.json
 FROM node:14-alpine
 
 ARG VERSION
+ARG REPOSITORY=directus/directus
 
 LABEL directus.version="${VERSION}"
+LABEL org.opencontainers.image.source https://github.com/${REPOSITORY}
 
 ENV \
-  PORT=41201 \
+  PORT="41201" \
   PUBLIC_URL="/" \
   DB_CLIENT="sqlite3" \
-  DB_FILENAME="/directus/database/data.db" \
+  DB_FILENAME="/directus/database/database.sqlite" \
   RATE_LIMITER_ENABLED="false" \
   RATE_LIMITER_STORE="memory" \
   RATE_LIMITER_POINTS="25" \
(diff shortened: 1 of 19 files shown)


.github/actions/build-images/rootfs/directus/images/main/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+SHELL ["/bin/bash", "-c"]
+
+WORKDIR /directus
+
+# Global requirements
+RUN npm install -g yargs pino pino-colada
+
+# Install Directus
+COPY --from=builder /directus/package.json .
```
nachogarcia, line 70: Would it make sense to copy also `package-lock.json` and run `npm ci` instead of install to keep exact dependency versions?

Also, I don't know if it would be better to use a different user instead of root, making the copy with `--chown=node:node` so that later on you can make use of `USER node`.

WoLfulus: I'm not sure about package lock, because what I am emulating here is the project creation thing when you `yarn create directus-project`. It basically installs directus as a dependency and then install the additional dependencies based on the database vendor. So it would always be the latest version of the package.

Did I miss something @rijkvanzanten ?

rijkvanzanten: You're right. It just installs directus, which has its own package lock downstream

WoLfulus: I can't create it through the CLI because of how it asks for inputs, so I just created a package.json file with all the database vendors bundled and I'm just updating the directus version before installing.

WoLfulus:
This should be ready for review. Please check the signal processing stuff on `start command` since I did not test it here.

Makefile
```diff
@@ -0,0 +1,30 @@
+SHELL=bash
+
+version=v9.0.0-beta.1
```
rijkvanzanten, line 3: Do we have to update this manually?

WoLfulus: The makefile is there for testing purposes only. This is just the default value, but you can change it with `make command version=v10.5.3`

rijkvanzanten (approved):

PR #415, opened 2020-09-24
Tweak shortcuts

### LINKED ISSUE #235: Cleanup Keyboard Shortcuts
We need to do a thorough audit of this throughout the system to ensure things work as expected (system + Directus). Here are some key tasks:

* [ ] Don’t override interface keyboard (eg: CMD+A within input)
* [ ] Meta-S to save
* [ ] Meta-Enter to submit a comment
* [ ] Escape should close any modal (if close is allowed)

Nitwel:
closes #235
Shortcuts now allow for multiple handlers on a single event.
Its also allows us to close any modal on `escape` and in the right order!
GIF 24 09 2020 11-22-02


app/src/components/v-dialog/v-dialog.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="v-dialog">
+	<div class="v-dialog" ref="dialog">
```
rijkvanzanten, line 2: I don't think this ref is used

app/src/components/v-dialog/v-dialog.vue
```diff
@@ -31,6 +32,18 @@ export default defineComponent({
...
 	setup(props, { emit }) {
+		const dialog = ref<HTMLElement | null>(null);
+		useShortcut(
+			() => {
+				if (_active.value) {
+					emitToggle();
+					return true;
+				}
+			},
+			ref(document.body),
```
rijkvanzanten, line 43: Can we make `ref(document.body)` the default?

app/src/components/v-table/v-table.vue
```diff
@@ -292,9 +294,13 @@ export default defineComponent({
...
-		useShortcut('mod+a', () => {
-			onToggleSelectAll(!allItemsSelected.value);
-		});
+		useShortcut(
+			() => {
+				onToggleSelectAll(!allItemsSelected.value);
+			},
+			table,
+			'meta+a'
+		);
```
rijkvanzanten, line 303: Lets remove the shortcut from the table, in favor of doing the shortcut in the tabular layout

app/src/composables/use-shortcut/readme.md
```diff
@@ -17,7 +17,7 @@ import { useShortcut } from '@/composables/use-shortcut';
 
 export default defineComponent({
 	setup(props) {
-		useShortcut('mod+s', save);
+		useShortcut('meta+s', save);
```
rijkvanzanten, line 20: This uses a different order (the correct one) than the actual use

app/src/composables/use-shortcut/use-shortcut.ts
```diff
@@ -1,27 +1,91 @@
...
+}
+
+function callHandlers(event: KeyboardEvent) {
+	Object.entries(handlers).forEach(([key, value]) => {
+		const rest = key.split('+').filter((keySegment) => keysdown.has(keySegment) === false);
+		if (rest.length > 0) return;
+		for (let i = 0; i < value.length; i++) {
+			const cancel = value[i](event);
+
+			// if the return value is true discontinue going through the queue.
```
rijkvanzanten, line 47: So `return` false at the end of the shortcut _stops_ it from firing the next?

rijkvanzanten: If yes, lets reverse that to be 

```
function handler() {
  return false;
}
```

to cancel out the next one

app/src/composables/use-shortcut/use-shortcut.ts
```diff
@@ -1,27 +1,91 @@
...
+		}
+	});
+}
 
 export default function useShortcut(
-	shortcut: string | string[],
-	handler: (evt?: ExtendedKeyboardEvent, combo?: string) => void
+	handler: ShortcutHandler,
+	reference: Ref<HTMLElement | null> | Ref<Vue | null>,
+	...shortcuts: string[]
```
rijkvanzanten, line 56: Lets make this `useShortcut(shortcut: (string | string[]), handler: ShortcutHandler, reference = ref(document.body))`

app/src/composables/use-shortcut/use-shortcut.ts
```diff
@@ -1,27 +1,91 @@
...
 ) {
+	const callback: ShortcutHandler = (event) => {
+		if (reference.value === null) return;
```
rijkvanzanten, line 59: Lets default the reference el to document.body

app/src/modules/settings/routes/data-model/field-detail/field-detail.vue
```diff
@@ -112,11 +113,14 @@ export default defineComponent({
...
 		type: {
-			type: String as PropType<'standard' | 'file' | 'files' | 'm2o' | 'o2m' | 'm2m' | 'presentation' | 'translations'>,
+			type: String as PropType<
+				'standard' | 'file' | 'files' | 'm2o' | 'o2m' | 'm2m' | 'presentation' | 'translations'
+			>,
 			default: null,
 		},
 	},
 	setup(props) {
+		const active = ref(true);
```
rijkvanzanten, line 123: I don't think we have to track `active` in here. We can leave `active` up top (L3) to be `true`, and just add that `@toggle` handler

rijkvanzanten (changes requested):

app/src/composables/use-shortcut/use-shortcut.ts
Nitwel: I'll leave that in to make sure that an existing element is passed over.

rijkvanzanten (approved):

PR #526, opened 2020-10-05
Update of API specifications.

thulioprado:


packages/spec/specs/paths/settings/setting.yaml
```diff
@@ -2,28 +2,28 @@ get:
...
-  - Settings
+    - Settings
 patch:
   description: Update an existing setting
   operationId: updateSetting
   parameters:
-    - $ref: '../../openapi.yaml#/components/parameters/Meta'
+    - $ref: "#/components/parameters/Meta"
   requestBody:
     content:
```
Nitwel, line 28: Can we also use a reference here or doesn't that match the setting schema?

packages/spec/specs/paths/settings/settings.yaml
```diff
@@ -32,26 +32,26 @@ patch:
         schema:
```
Nitwel, line 32: This doesn't seem to be up to date.

packages/spec/specs/paths/webhooks/webhook.yaml
```diff
@@ -0,0 +1,87 @@
...
+patch:
+  description: Update an existing webhook
+  operationId: updateWebhook
+  parameters:
+    - $ref: "#/components/parameters/Fields"
+    - $ref: "#/components/parameters/Meta"
+  requestBody:
+    content:
+      application/json:
+        schema:
```
Nitwel, line 29: we shoud use a reference here.

packages/spec/specs/paths/webhooks/webhooks.yaml
```diff
@@ -2,18 +2,72 @@ get:
...
+post:
+  description: Create a new webhook.
+  operationId: createWebhook
+  parameters:
+    - $ref: "#/components/parameters/Fields"
+    - $ref: "#/components/parameters/Meta"
+  requestBody:
+    content:
+      application/json:
+        schema:
```
Nitwel, line 29: Should use reference.

Nitwel (changes requested):
This is looking pretty good. 👍 
There are some small changes to be made but other than that well done.

Something we should change is that we use references in requestBodys rather than having dublicated schemas.

We also need to look at whats required in a schema because I think there might be missing some required fields.

Also,how do we want to store examples.
We could either store them inside the root object and have one big example or have an example for each property of that object. We could also do both.

Was there a reason for replacing `../../openapi.yaml`?

rijkvanzanten:
And last but not least; lets resolve the merge conflicts

PR #547, opened 2020-10-06
Fix back button opening modal

### LINKED ISSUE #545: Back arrow opens modal instead of going back to the data-modal view.
grafik

Pressing the arrow should lead back to the main data-model page. When having edited a field the last url becomes the field so instead of going back the field modal opens up again.

Nitwel:
closes #545

### DIFF
--- a/app/src/modules/collections/routes/detail.vue
+++ b/app/src/modules/collections/routes/detail.vue
@@ -33,7 +33,7 @@
 				secondary
 				exact
 				v-tooltip.bottom="$t('back')"
-				:to="'/collections/'+collection"
+				:to="'/collections/' + collection"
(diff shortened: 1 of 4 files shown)


app/src/modules/collections/routes/detail.vue
```diff
@@ -33,7 +33,7 @@
 				secondary
 				exact
 				v-tooltip.bottom="$t('back')"
-				@click="$router.go(-1)"
+				:to="'/collections/'+collection"
```
rijkvanzanten, line 36:
```suggestion
				:to="'/collections/' + collection"
```

rijkvanzanten (approved):

PR #575, opened 2020-10-07
Fix opening repeater row when creating new one

### LINKED ISSUE #574: Repeater not properly opening when adding a new item
9.0.0beta7 docker image with DB pointing on a mysql 5.7

Issue: When adding a new item of a repeater, the repeater adds a new item in a closed state, and when clicking to open it, it opens and closes right away. Only the next click finally properly opens it.

Expected behavior: Adding a new item of a repeater adds the new item in an open state.

Nitwel:
closes #574

### DIFF
--- a/app/src/components/v-item-group/v-item.vue
+++ b/app/src/components/v-item-group/v-item.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="v-item">
-		<slot v-bind="{ 'active': isActive, toggle }" />
+		<slot v-bind="{ active: isActive, toggle }" />
 	</div>
 </template>
 
--- a/app/src/components/v-item-group/v-item.vue
+++ b/app/src/components/v-item-group/v-item.vue
@@ -1,11 +1,11 @@
 <template>
 	<div class="v-item">
-		<slot v-bind="{ active, toggle }" />
+		<slot v-bind="{ 'active': isActive, toggle }" />
 	</div>
 </template>
 
 <script lang="ts">
-import { defineComponent } from '@vue/composition-api';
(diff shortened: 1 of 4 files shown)


app/src/components/v-item-group/v-item.vue
```diff
@@ -1,11 +1,11 @@
 <template>
 	<div class="v-item">
-		<slot v-bind="{ active, toggle }" />
+		<slot v-bind="{ 'active': isActive, toggle }" />
```
rijkvanzanten, line 3:
```suggestion
		<slot v-bind="{ active: isActive, toggle }" />
```

app/src/components/v-item-group/v-item.vue
```diff
@@ -18,14 +18,25 @@ export default defineComponent({
...
 			default: 'item-group',
 		},
+		active: {
+			type: Boolean,
+			default: undefined
+		},
+		watch: {
+			type: Boolean,
+			default: true
+		}
```
rijkvanzanten, line 28: What does watch do

rijkvanzanten (changes requested):

app/src/components/v-item-group/v-item.vue
Nitwel: enables/disables the watcher. If set to false the `active ref<bool>` will not update the state of the item after it got initialized.

rijkvanzanten: Could you elaborate on that? I don't know what you mean by 

> If set to false the active ref<bool> will not update the state of the item after it got initialized.

rijkvanzanten (approved):

PR #580, opened 2020-10-08
Add relation between presets and roles to seeds

nickrum:
This fixes the presets and bookmarks browse view showing All instead of
the role if the scope is set to a role.


rijkvanzanten (approved):

rijkvanzanten:
Appreciate the help @nickrum!

PR #619, opened 2020-10-12
Relational updates

Nitwel:
- [x] Update o2m
- [x] Update m2m


rijkvanzanten (changes requested):
Noticing a couple things:

- [ ] Interfaces try fetching existing items, even if value is null or empty array
- [ ] Creating a new item on m2m doesn't work
- [ ] Selecting existing item on m2m doesn't work
   (both try fetching with `filter[id][_in]=` with no value
- [ ] Can't create new item on o2m ("cannot read property map of null" in stageEdits)
- [ ] Random things show up in console at times (couple stray console.log's it seems)

rijkvanzanten (approved):

PR #620, opened 2020-10-12
Fix docs navigation

Nitwel:
The navigation would fail to render because the `.bin` directory made it into folders.
It also would fail if props.path was null.


rijkvanzanten:
Shouldn't the solution then be to ensure that `.bin` is ignored?

Nitwel:
That too, I just couldn't find how this even got in there because it is not in the json file.

rijkvanzanten (approved):

Issue #363, opened 2020-09-21
CLI Init fails on SQL server due to port number / string issue


kukulaka:
**Directus**: 9.0.0-beta.1
**Node**:  v12.18.3
**SQL Server**: 15.0.2000.5

Initialising CLI

knex / tedious error 

```
(node:21392) UnhandledPromiseRejectionWarning: TypeError: The "config.options.port" property must be of type number.
    at new Connection (C:\Users\kukul\Documents\next\api\node_modules\tedious\lib\connection.js:620:17)
    at C:\Users\kukul\Documents\next\api\node_modules\mssql\lib\tedious\connection-pool.js:61:23

"config.options.port" property must be of type number.


at Client_MSSQL.acquireRawConnection (C:\Users\kukul\Documents\next\api\node_modules\knex\lib\dialects\mssql\index.js:221:12)
```

This is specific to MSSQL driver. 1433 was entered as port. If port is not entered this error is not thrown. 
``

rijkvanzanten:
What did you use for the env var @kukulaka ? `DB_PORT="1433"` or `DB_PORT=1433` ?

kukulaka:
@rijkvanzanten DB_PORT=1433

mscbpi:
@rijkvanzanten the same issue stills appear if you enter a port from default when installing with `npx` (CLI wizard). 

Works when let empty (default). Would that mean that if `DB_PORT` is not set as env with `DB_CLIENT` mssql, it would default to 1433 ?

PR #623, opened 2020-10-12
Convert Numeric Values of Environment Variables to Number

mscbpi:
convert numeric values to number

Closes https://github.com/directus/next/issues/363

### DIFF
--- a/api/src/env.ts
+++ b/api/src/env.ts
@@ -62,6 +62,7 @@ function processValues(env: Record<string, any>) {
 		if (value === 'true') env[key] = true;
 		if (value === 'false') env[key] = false;
 		if (value === 'null') env[key] = null;
+		if (isNumeric(value)) env[key] = Number(value);
 	}
 
 	return env;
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
@mscbpi That `isNumeric` function doesn't seem to exist in this file scope.. You might wanna check using `isNaN` or add/import that `isNumeric` function to this file

mscbpi:
> @mscbpi That `isNumeric` function doesn't seem to exist in this file scope.. You might wanna check using `isNaN` or add/import that `isNumeric` function to this file

my bad.

api/src/env.ts
```diff
@@ -62,7 +62,7 @@ function processValues(env: Record<string, any>) {
 		if (value === 'true') env[key] = true;
 		if (value === 'false') env[key] = false;
 		if (value === 'null') env[key] = null;
-		if (isNumeric(value)) env[key] = Number(value);
+		if (!isNaN(Number(value))) env[key] = Number(value);
```
rijkvanzanten, line 65:
```suggestion
		if (isNaN(value) === false) env[key] = Number(value);
```

rijkvanzanten (changes requested):
AFAIK `isNaN` accepts any type for it's parameter, eg `isNaN('12') == false` and `isNaN('rijk') == true`.

Lets update the check to be a little more explicit

rijkvanzanten (approved):

mscbpi:
With typescript it needs to be typed.

PR #627, opened 2020-10-12
Add project icon (#429)

### LINKED ISSUE #429: Use "Project Logo" as favicon
This could either be an option or a replacement for the current color "blob".

The project logo makes more sense over just a color.

maartenvn:
This adds an option to the settings to change the project favicon.

Standard behavior of the favicon has not been changed, when the `project_icon` setting is set, the icon will be used instead of the default color dot.

In order to add the setting, the project settings layout has been altered:

image

closes #429


benhaynes:
Thanks @maartenvn!! I love the idea, but I don't think this passes 80/20 specifically in the App. I would like to find a way to add this feature, but don't want to confuse the Settings in the App with another "project icon".

PR #686, opened 2020-10-16
Add sorting to m2m and o2m

### LINKED ISSUE #360: Add a "Enable Sorting" option to M2M/O2M Setup
It is a common use-case to have sorting on O2M and M2M relationships. Adding this to the setup flow will reduce confusion on how to accomplish this.

We'll use the "toggle/value" component for this... the only bummer is that it will increase the height of the relational setup and make it seem more confusing.

Nitwel:
closes #360


rijkvanzanten:
When firing up the app, I'm greeted with a TS error

CleanShot 2020-10-20 at 15 12 59

Could you make sure the files interface is updated too?

rijkvanzanten (approved):

PR #693, opened 2020-10-17
Fix translation not set before initial load

### LINKED ISSUE #692: Translations for collections not set on initial load
**Directus**: 9.0.0-rc.0
**MySQL**: Ver 14.14

## Reproduce
1. Create a collection
2. Update the translation for the language set for your current user
3. Go to the collection overview page (everything works fine)
4. Reload that page (`/admin/collections/`)
5. Collections are now displayed in the dotted notation rather than the translated value

Navigating to a different page & going back without refreshing will display the correct translation again

maartenvn:
`setLanguage` wasn't correctly awaited when hydrating the app. This explains why the bug was so random, because it would show the incorrect translation only when it didn't resolve in time before the next promise :smile: 

Maybe adding an eslint check to await promises/handle promise result could prevent something like this in the future? (eg: `@typescript-eslint/no-floating-promises`)

close #692

### DIFF
--- a/app/src/hydrate.ts
+++ b/app/src/hydrate.ts
@@ -57,7 +57,7 @@ export async function hydrate(stores = useStores()) {
 		 */
 		await userStore.hydrate();
 
-		setLanguage((userStore.state.currentUser?.language as Language) || 'en-US');
+		await setLanguage((userStore.state.currentUser?.language as Language) || 'en-US');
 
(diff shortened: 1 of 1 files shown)


benhaynes:
Thank you!! We'll review ASAP... 👍

rijkvanzanten (approved):

PR #701, opened 2020-10-17
Some form field menu improvements

nickrum:
The first two commits are just quality of life changes.

The third commit also prevents a server error when a toggle interface is cleared and submitted.
More specifically this is a "SQLITE_CONSTRAINT: NOT NULL constraint failed" error, but I think the clear option should not be shown in general for fields that are set to be required and have a default value.

### DIFF
--- a/app/src/components/v-form/form-field-menu.vue
+++ b/app/src/components/v-form/form-field-menu.vue
@@ -61,8 +61,7 @@ export default defineComponent({
 		});
 
 		const isRequired = computed(() => {
-			const isNullable = props.field?.schema?.is_nullable;
-			return !isNullable;
+			return props.field?.schema?.is_nullable === false;
 		});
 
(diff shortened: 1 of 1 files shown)


app/src/components/v-form/form-field-menu.vue
```diff
@@ -52,7 +60,12 @@ export default defineComponent({
 			return savedValue !== undefined ? savedValue : null;
 		});
 
-		return { defaultValue };
+		const isRequired = computed(() => {
+			const isNullable = props.field?.schema?.is_nullable;
+			return !isNullable;
+		});
```
rijkvanzanten, line 66:
```suggestion
		const isRequired = computed(() => {
			return props.field?.schema?.is_nullable === false;
		});
```

rijkvanzanten (approved):

Issue #705, opened 2020-10-18
Custom asset presets don't work


dimitrov-adrian:
Custom presets doesn't work even when set to All and give "Dynamic asset generation has been disabled for this project."

PR #706, opened 2020-10-18
Assets fixes and improvements.

dimitrov-adrian:
- Fix https://github.com/directus/next/discussions/474
image

- Fix https://github.com/directus/next/issues/705
- Adding inside and outside fit options

But there is a downside with the shorthands `w`, `h`, `f` which are renamed to `width`, `height` and `fit` as they where lost with the transformation object on https://github.com/directus/next/compare/main...dimitrov-adrian:assets?expand=1#diff-ce9f53f0d1bc147c6901819c7e0ab71063b3a2005db78706e66ab4fd71368537R82 But I had update the documentation file as well.


api/src/controllers/assets.ts
```diff
@@ -118,6 +117,7 @@ router.get(
 
 		res.attachment(file.filename_download);
 		res.setHeader('Content-Type', file.type);
+		res.removeHeader('Content-Disposition');
```
rijkvanzanten, line 120: Why are you explicitly removing the content disposition header here? That's needed to download the file when used directly in the browser

rijkvanzanten (changes requested):
Lets keep the naming consistent with `sharp`, and use `without-enlargement` as the query param, instead of `noupscale`

rijkvanzanten:
The fork of this PR has been deleted. We'll continue this on https://github.com/directus/next/pull/787

PR #731, opened 2020-10-20
Add user invitation

### LINKED ISSUE #119: Add "User Invite" to Create User Flow
Dedicated button on Settings > Roles > Users ... "tags" style interface that is a CSV of email addresses that receive invites via email.

It would be nice to also have this flow integrated into the User Browse/Detail page too, perhaps as a dedicated "Invite User" action button in the header next to "New User".

Nitwel:
closes #119 
grafik
grafik

### DIFF
--- a/app/src/views/private/components/users-invite/users-invite.vue
+++ b/app/src/views/private/components/users-invite/users-invite.vue
@@ -5,7 +5,7 @@
 
 			<v-card-text class="grid">
 				<div class="field">
-					<span class="type-label">{{ $t('emails') }}</span>
+					<div class="type-label">{{ $t('emails') }}</div>
 					<interface-tags
 						v-model="emails"
 						:placeholder="$t('add_to_invite_user')"
@@ -14,7 +14,7 @@
 					></interface-tags>
 				</div>
 				<div class="field" v-if="role === null">
-					<span class="type-label">{{ $t('role') }}</span>
+					<div class="type-label">{{ $t('role') }}</div>
 					<v-select v-model="roleSelected" :items="roles"></v-select>
 				</div>
 			</v-card-text>
--- a/app/src/views/private/components/users-invite/users-invite.vue
+++ b/app/src/views/private/components/users-invite/users-invite.vue
@@ -106,7 +106,7 @@ export default defineComponent({
 });
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
 @import '@/styles/mixins/form-grid';
 
 .v-card-text {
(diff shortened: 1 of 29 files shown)


benhaynes:
Perfect!! Thank you Nils...

Two questions:
* Let's ensure the proper label/field spacing is there... I think it's 8px, but we should confirm.
* Should this be a textarea instead of an input? Might make sense for longer CSVs

Nitwel:
It is using the tags interface. So the button only gets clickable when you added a tag containing an email.
The default spacing for that is 4px as far as I know but I can look into that.

benhaynes:
Confirmed that it's 8px (see below). Ahh, tags interface works!

Screen Shot 2020-10-20 at 10 00 58 AM

Nitwel:
The `type-label` was a span and not a div. That's why it ignored that margin.

app/src/views/private/components/users-invite/users-invite.vue
```diff
@@ -0,0 +1,121 @@
...
+
+			<v-card-text class="grid">
+				<div class="field">
+					<div class="type-label">{{ $t('emails') }}</div>
+					<interface-tags
+						v-model="emails"
+						:placeholder="$t('add_to_invite_user')"
+						icon-right="email"
+						whitespace=""
+					></interface-tags>
```
rijkvanzanten, line 14:
```suggestion
					/>
```

app/src/views/private/components/users-invite/users-invite.vue
```diff
@@ -0,0 +1,121 @@
...
+				</div>
+				<div class="field" v-if="role === null">
+					<div class="type-label">{{ $t('role') }}</div>
+					<v-select v-model="roleSelected" :items="roles"></v-select>
```
rijkvanzanten, line 18:
```suggestion
					<v-select v-model="roleSelected" :items="roles" />
```

rijkvanzanten (changes requested):

app/src/views/private/components/users-invite/users-invite.vue
```diff
@@ -0,0 +1,121 @@
...
+					persist: true,
+					type: 'error',
+				});
+			} finally {
+				loading.value = false;
+			}
+		}
+
+		async function loadRoles() {
+			const response = await api.get('/roles');
```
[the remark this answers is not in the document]
Nitwel: How would I check for that?

rijkvanzanten (approved):

PR #734, opened 2020-10-20
Allow for custom formats

### LINKED ISSUE #211: Let users format date/ datetime string
Hi, currently date/ DateTime fields rely on user locale when rendering values.

Depending on a locale, this can take too much space in the table. It would be great to allow users to choose if they need long string or short string, or even let to define the format manually, such as yyyy-mm-dd HH:MM:ss as in https://momentjs.com/docs/#/displaying/

image

Nitwel:
closes #211 
It is now possible to choose between `Long`, `Short` and `Other` on the datetime display.

### DIFF
--- a/app/src/lang/en-US/index.json
+++ b/app/src/lang/en-US/index.json
@@ -390,7 +390,7 @@
 	"date-fns_date": "PPP",
 	"date-fns_time": "h:mma",
 	"date-fns_date_short": "MMM d, u",
-	"date-fns_time_short": "hh:mm",
+	"date-fns_time_short": "h:mma",
 	"date-fns_date_short_no_year": "MMM d",
 	"month": "Month",
 	"year": "Year",
--- a/app/src/lang/en-US/index.json
+++ b/app/src/lang/en-US/index.json
@@ -390,6 +390,7 @@
 	"date-fns_date": "PPP",
 	"date-fns_time": "h:mma",
 	"date-fns_date_short": "MMM d, u",
+	"date-fns_time_short": "hh:mm",
 	"date-fns_date_short_no_year": "MMM d",
 	"month": "Month",
 	"year": "Year",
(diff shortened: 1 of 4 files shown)


rijkvanzanten (approved):

app/src/lang/en-US/index.json
```diff
@@ -390,6 +390,7 @@
 	"date-fns_date": "PPP",
 	"date-fns_time": "h:mma",
 	"date-fns_date_short": "MMM d, u",
+	"date-fns_time_short": "hh:mm",
```
rijkvanzanten, line 393: @Nitwel This should be including am/pm, `en-US` uses 12h time 🙂

rijkvanzanten (changes requested):

app/src/lang/en-US/index.json
```diff
@@ -390,7 +390,7 @@
 	"date-fns_date": "PPP",
 	"date-fns_time": "h:mma",
 	"date-fns_date_short": "MMM d, u",
-	"date-fns_time_short": "hh:mm",
+	"date-fns_time_short": "h:mma",
```
rijkvanzanten, line 393: But now it's exactly the same as regular `fns_time` 😁 What's the planned difference between "time" and "short time" ?

rijkvanzanten (changes requested):

app/src/lang/en-US/index.json
Nitwel: Should we include seconds in long time?
I differentiated so other languages could differentiate between them.

rijkvanzanten: That works!

benhaynes:
Is "Relative Time" (eg: `3d ago` or `6 weeks ago`) an option here, elsewhere, or not at all?

Nitwel:
It is a separate toggle which completely ignores any format setting if set to true. We could also add this option into the dropdown but I do like how it is now.

benhaynes:
Perfect — how it is now works for me!

PR #758, opened 2020-10-22
Remove outdated required field property

nickrum:
@rijkvanzanten is the `FieldMeta.required` property still used anywhere? Otherwise, I think it can be removed here and in a few other places.


Nitwel:
`meta.required` is outdated and can be removed.

nickrum:
Thanks @Nitwel. I updated the PR to remove the property.

rijkvanzanten (approved):

PR #759, opened 2020-10-22
Add error handling

### LINKED ISSUE #506: Show Notification for Async Actions
Some, if not all, async actions that are saved inline should show a notification so the user is aware something happened. 

## Show
* Changing a file's folder
* Adding a field
* Adding a collection

## Don't Show
* Reordering Fields

### LINKED ISSUE #685: No feedback on item creation
When there's a server violation (like duplicated unique values, non null values) when creating an item, the UI doesn't display any feedback and/or error.

Nitwel:
closes #506 closes #685


app/src/api.ts
```diff
@@ -79,6 +82,13 @@ export const onError = async (error: RequestError) => {
 		}
 	}
 
+	notify.add({
+		title: i18n.t(`errors.${code}`),
+		text: message,
+		type: 'error',
+		dialog: true,
+	});
+
```
rijkvanzanten, line 91: Lets not do it every failed request. Some requests are expected to "error" and are handled elsewhere. We should throw the notification in the `catch` where we try making requests

app/src/composables/use-item/use-item.ts
```diff
@@ -89,16 +90,16 @@ export function useItem(collection: Ref<string>, primaryKey: Ref<string | number
 			let response;
 
 			if (isNew.value === true) {
-				response = await api.post(endpoint.value, edits.value);
+				response = await api.post('A' + endpoint.value, edits.value);
```
rijkvanzanten, line 93: You probably broke this on purpose for debugging reasons, but lets not keep this in production 😉

app/src/composables/use-item/use-item.ts
```diff
@@ -134,7 +117,7 @@ export function useItem(collection: Ref<string>, primaryKey: Ref<string | number
 					});
 			}
 
-			throw err;
+			throw error;
```
rijkvanzanten, line 120: `error` doesn't exist here. It's `err`

app/src/composables/use-item/use-item.ts
```diff
@@ -157,30 +140,21 @@ export function useItem(collection: Ref<string>, primaryKey: Ref<string | number
...
 		} catch (err) {
-			notify({
-				title: i18n.t('item_create_failed'),
-				text: i18n.tc('item_in', isBatch.value ? 2 : 1, {
-					collection: collection.value,
-					primaryKey: isBatch.value ? (primaryKey.value as string).split(',').join(', ') : primaryKey.value,
-				}),
-				type: 'error',
-			});
-
```
rijkvanzanten, line 175: We should keep notifying in the catches here. Also, not every failed request should trigger the full "unexpected error dialog", only the errors that are truly unexpected.

app/src/composables/use-item/use-item.ts
```diff
@@ -157,30 +140,21 @@ export function useItem(collection: Ref<string>, primaryKey: Ref<string | number
...
 			if (err?.response?.data?.errors) {
 				validationErrors.value = err.response.data.errors
 					.filter((err: APIError) => err.extensions.code === 'FAILED_VALIDATION')
 					.map((err: APIError) => {
 						return err.extensions;
 					});
 			} else {
-				throw err;
+				throw error;
```
rijkvanzanten, line 157: `error` doesn't exist here

app/src/composables/use-item/use-item.ts
```diff
@@ -245,20 +210,11 @@ export function useItem(collection: Ref<string>, primaryKey: Ref<string | number
...
 		} catch (err) {
-			notify({
-				title: i18n.tc('item_delete_failed', isBatch.value ? 2 : 1),
-				text: i18n.tc('item_in', isBatch.value ? 2 : 1, {
-					collection: collection.value,
-					primaryKey: isBatch.value ? (primaryKey.value as string).split(',').join(', ') : primaryKey.value,
-				}),
-				type: 'error',
-			});
-
```
rijkvanzanten, line 261: See above

app/src/interfaces/file/file.vue
```diff
@@ -155,37 +155,33 @@ export default defineComponent({
...
 			urlLoading,
-			urlError,
 			importFromURL,
 			isValidURL,
 			assetURL,
 		};
 
 		function useFile() {
 			const loading = ref(false);
-			const error = ref(null);
```
rijkvanzanten, line 179: Why remove error here? We should have an error state for the file interface instead of just (silently) console logging the error

app/src/interfaces/file/file.vue
```diff
@@ -224,7 +220,6 @@ export default defineComponent({
 		function useURLImport() {
 			const url = ref('');
 			const loading = ref(false);
-			const error = ref(null);
```
rijkvanzanten, line 227: Keep / add error state here

app/src/interfaces/image/image.vue
```diff
@@ -84,7 +83,6 @@ export default defineComponent({
 	setup(props, { emit }) {
 		const loading = ref(false);
 		const image = ref<Image | null>(null);
-		const error = ref(null);
```
rijkvanzanten, line 87: Leave error state in here

rijkvanzanten (changes requested):
Couple things:

* The error dialog is a last-effort. Ideally, errors are handled gracefully within the app. The error dialog should only be used when there's no better way to handle errors (yet) in the component
* Interfaces etc that have an `error` ref should keep it. Those interfaces should handle errors gracefully and show a notice or something else in the interface itself instead of a big dialog. 
* We can't simply show the error on every non 2xx request, as we expect 400 / 401 / 403 / 404 in certain cases (eg trying to load a previously-deleted image shouldn't trigger a full dialog)
* We should never just `console.error(err)` in a catch block with no other logic. Those errors are invisible for the user, and are only visible to devs after telling them where to look
* The idea behind having an error ref is that the component itself can use that in the template (eg `<v-error v-if="error" :error="error" /> <p v-else>Component</p>` )
* ~~Lets rename `notify` to `notificationsStore` for consistency with other stores~~
* Why remove the `notify` util? Was nice to have it be that explicit, and prevent having to import the notificationsstore into all these components
* Likewise, I was under the impression we'd add a "`unexpectedError()`" (or something along those lines) util function

rijkvanzanten (approved):

PR #770, opened 2020-10-23
Allow set folder for new file(s) in field setup

dimitrov-adrian:
It resolved https://github.com/directus/next/discussions/621

1. Allow to set folder for file/files field in data model
2. API change, allowing to set folder when importing file from URL

UI

image

### DIFF
--- a/api/src/controllers/files.ts
+++ b/api/src/controllers/files.ts
@@ -132,6 +132,7 @@ router.post(
 
 const importSchema = Joi.object({
 	url: Joi.string().required(),
+	folder: Joi.string(),
 	data: Joi.object(),
 });
 
@@ -159,6 +160,7 @@ router.post(
 			type: fileResponse.headers['content-type'],
 			title: formatTitle(filename),
 			...(req.body.data || {}),
+			folder: req.body.folder,
 		};
 
 		const primaryKey = await service.upload(fileResponse.data, payload);
--- a/app/src/components/v-upload/v-upload.vue
+++ b/app/src/components/v-upload/v-upload.vue
@@ -116,6 +116,10 @@ export default defineComponent({
 			type: Boolean,
 			default: false,
 		},
+		folder: {
+			type: String,
+			default: null,
+		},
 	},
 	setup(props, { emit }) {
 		const { uploading, progress, error, upload, onBrowseSelect, done, numberOfFiles } = useUpload();
@@ -169,6 +173,7 @@ export default defineComponent({
 								done.value = percentage.filter((p) => p === 100).length;
 							},
 							preset: props.preset,
+							folder: props.folder,
 						});
 
 						uploadedFiles && emit('input', uploadedFiles);
@@ -180,6 +185,7 @@ export default defineComponent({
 							},
 							fileId: props.fileId,
 							preset: props.preset,
+							folder: props.folder,
 						});
 
 						uploadedFile && emit('input', uploadedFile);
@@ -274,6 +280,7 @@ export default defineComponent({
 				try {
 					const response = await api.post(`/files/import`, {
 						url: url.value,
+						folder: props.folder,
 					});
 
 					emit('input', response.data.data);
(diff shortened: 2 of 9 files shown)


benhaynes:
Nice idea! Love it...

Nitwel:
This is a great idea. But I think it would be better to have this option in the file/s and image interface rather than in the data-modal.
To allow this we could add a folder interface which allows you to select any folder in directus. This interface would be a system interface which means it would sit under `src/interfaces/_system`. @rijkvanzanten What do you think on this?

dimitrov-adrian:
Yes, other component for picking directories will be better in cases where tones of folders are created. Moving the component itself is not a problem since i had used options which are intended to be mostly for interfaces, but I didn't be sure which is more natural way to set such option. I think it it's related to data model where it stores the data, but it could be the interface too since it's the way user interact.

Nitwel:
Just hit me up in discord if you got questions on how interfaces work!

api/src/controllers/files.ts
```diff
@@ -132,6 +132,7 @@ router.post(
 
 const importSchema = Joi.object({
 	url: Joi.string().required(),
+	folder: Joi.string(),
```
rijkvanzanten, line 135: No need to add folder here. The full post body is already included under `data`, so any field (including `folder`) can be set

app/src/components/v-upload/v-upload.vue
```diff
@@ -180,6 +185,7 @@ export default defineComponent({
 							},
 							fileId: props.fileId,
 							preset: props.preset,
```
rijkvanzanten, line 187: Preset contains any fields on the file, including folder. No need to add it separately

app/src/interfaces/_system/folder/folder.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+	<folder-picker :value="currentFolder" :disabled-folders="disabledFolders" @input="onFolderSelect" />
```
rijkvanzanten, line 2: Instead of having a duplicate local state, you can use the `value` prop here directly

```suggestion
	<folder-picker :value="value" :disabled-folders="disabledFolders" @input="onFolderSelect" />
```

app/src/interfaces/file/file.vue
```diff
@@ -71,7 +71,7 @@
 			<v-card>
 				<v-card-title>{{ $t('upload_from_device') }}</v-card-title>
 				<v-card-text>
-					<v-upload @input="onUpload" from-url />
+					<v-upload @input="onUpload" :folder="$attrs.folder" from-url />
```
rijkvanzanten, line 74: Add `folder` as a prop instead of reading it from `$attrs`

app/src/interfaces/file/file.vue
```diff
@@ -243,6 +243,7 @@ export default defineComponent({
 				try {
 					const response = await api.post(`/files/import`, {
 						url: url.value,
+						folder: attrs.folder,
```
rijkvanzanten, line 246: Nest this under `data`

app/src/utils/upload-file/upload-file.ts
```diff
@@ -23,6 +24,10 @@ export default async function uploadFile(
 		}
 	}
 
+	if (options?.folder) {
+		formData.append('folder', options.folder);
+	}
```
rijkvanzanten, line 29: Everything under `options.preset` is already included in the formData. No need to add `folder` separately

rijkvanzanten (changes requested):
See above

api/src/controllers/files.ts
dimitrov-adrian: If don't have this I got error that folder is not allowed

rijkvanzanten: Should be fine if you nest `folder` under `data` in the payload, so:

```json
{
  "url": "example.com/image.png",
  "data": {
    "folder": "abcdef"
  }
}
```

dimitrov-adrian:
I have changed the folder component selector as @Nitwel requested to own (dropdown)

image

PR #775, opened 2020-10-24
Add generics for ItemsService

apstanisic:
Adds ability for extension developer to specify type returned from ItemsService. This helps have better
intellisense and better type safety when developing.

I imported `Item` as `AnyItem` because that way code stays more readable.
And I had to specify types in some places because other services don't have generic types.

Now we can do:
```typescript
interface Post {
  name: string;
  id: number;
}

const posts = new ItemsService<Post>('posts');
await posts.update({invalidKey: 'value'}, 1); // It will show error in IDE
const firstPost = await posts.readByQuery({ filter: { id: 1 }, single: true }); // Returns Post
```

### DIFF
--- a/api/src/services/items.ts
+++ b/api/src/services/items.ts
@@ -26,7 +26,7 @@ import getDefaultValue from '../utils/get-default-value';
 import { InvalidPayloadException } from '../exceptions';
 import { ForbiddenException } from '../exceptions';
 
-export class ItemsService<Item extends AnyItem> implements AbstractService {
+export class ItemsService<Item extends AnyItem = AnyItem> implements AbstractService {
 	collection: string;
 	knex: Knex;
 	accountability: Accountability | null;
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
Thanks @apstanisic! One thing to keep in mind here is that the shape of the object can/will be different between create/read/update/delete per user, so a single generic for the whole class may or may not work as expected based on your use case. Can we support overriding on the individual methods as well?

rijkvanzanten:
Thanks @apstanisic! Final question; any reason why / why not to do it for `readSingleton` and/or the other services? (eg ActivityService / PermissionsService etc)

apstanisic:
I've added this option, but the code is more verbose. Should I just remove generic type from class and require user to specify type for every method if he/she wants better types? Also it does not provide full type safety because TypeScript will infer value from param, and will replace type in class constructor.

```typescript
const items = new ItemsService<{id: number}>('items');

// Code bellow will work, ts will infer T as {name: string}, and replace type from constructor
items.update({name: 'Test'}, 1); 
```

rijkvanzanten:
Yeah,, that makes sense to me 👍

apstanisic:
There was no return  type for readSingleton, so I was not sure if there is any special type returned. I'll add it. 
I started with ItemsService because I was sure in types that it returns, I'll add more services when I read the code better.

apstanisic:
I've added to `readSingleton`.
Should I remove generics from constructor, I'm not sure how confusing it is for developers that they can pass value to methods that does not satisfy type in constructor? Most people expect methods to be type safe, and this is not

rijkvanzanten:
Right, the "downside" of TS in the context of Directus is the fact that the IO of the methods/classes is based on the authenticated user and configured schema, so we'll never achieve 100% static code type safety 🤔

apstanisic:
We can always return/accept Partial<T>, that way user can specify all fields, and then force him to check if field is null, but it's not best DX

rijkvanzanten:
I think that's the cleanest solution for now. Worst case scenario, you can always cast things with `await service.create(...) as MyType;`

apstanisic:
`service.create/update/upsert/delete` always return primary key, so that's not a problem. They also accept partial, because pk and some fields should not be passed. The only methods that user needs to specify are `read` methods, and there is no way to know, it will always depend on permissions

rijkvanzanten (approved):

PR #777, opened 2020-10-25
Add editor.js (block editor) interface

dimitrov-adrian:


maartenvn:
Can we get a demo of this :slightly_smiling_face:?
I like the idea of having this interface, however I am not sure if it should be included inside Directus. It adds an additional 349.6
kB (for just `@editorjs/editorjs`) to the bundle size of the main app, while not everyone will use this future. It is a high price to pay to have this included

https://bundlephobia.com/result?p=@editorjs/editorjs@2.19.0

dimitrov-adrian:
@maartenvn sure, I am attaching some screenshots

Screen Shot 2020-10-25 at 20 17 16
Screen Shot 2020-10-25 at 20 17 31
Screen Shot 2020-10-25 at 20 18 55
Screen Shot 2020-10-25 at 20 16 57

For see it in action you could just pull the branch of the PR.

About the weight that it will add, you are right, it really add some weight to the client (even that the size is bit smaller when gziped and is considered to be cached) for something that is sure will not be used by 100% of users, BUT, for sure there is a lot of people that do not use markdown editor, code editor and even the HTML wysiwyg, is this means that they should be removed?

I think in one perfect world (or hopeful in near future) built in interfaces should be cutted to very few and basic just like the HTML inputs (even the tags and color should be considered as a fancy components) and everything else will be available from some kind of marketplace.

Anyway, I really get your point of view, thanks for the comment.

rijkvanzanten:
Re: weight, we can do a similar trick as in the code editor and only (async) load in the editor and it's plugins when they're actually used, instead of all the time. 

> I think in one perfect world (or hopeful in near future) built in interfaces should be cutted to very few and basic just like the HTML inputs (even the tags and color should be considered as a fancy components) and everything else will be available from some kind of marketplace.

I 💯 agree with this. 

---

I have a couple follow up questions:

1. Is there still a need for the current WYSIWYG interface with this one included?
2. How does this save to the DB? JSON?
  2b. If yes, is there a way to save the parts relationally, so it's better indexable/searchable etc?
3. How does it integrate with files?

benhaynes:
We'd have to do some serious testing and ensure that all basic WYSIWYG features are preset before replacing the current WYSIWYG. I know HTML is not "agnostic" content, but I do think that many users are still looking for an HTML output editor in Core. Rijk has asked some key questions above, so let's start there.

The async loading should hopefully resolve the size issue, and it will be really nice when the built-in marketplace is done and polished up, so that can be the main way to add/remove/manage interfaces! 😄

maartenvn:
>     1. Is there still a need for the current WYSIWYG interface with this one included?

Yes, very much for myself & my team. We need raw HTML as an output

apstanisic:
> 1. Is there still a need for the current WYSIWYG interface with this one included?

I think there is. 
WYSIWYG returns HTML that is easy to show on Android and iOS, app can simply fetch data and show it in a webview, while for editorjs, it needs to be server rendered to be shown on phones. If it's server rendered it also means that we must fetch data trough Directus, and not from db directly.

I think for most people should use WYSIWYG for blogs, while editorjs for page building.

dimitrov-adrian:
> I have a couple follow up questions:
> 
> 1. Is there still a need for the current WYSIWYG interface with this one included?
Yes and no, WYSIWYG is most commonly used for web pages, while block editor output JSON

> 2. How does this save to the DB? JSON?
>    2b. If yes, is there a way to save the parts relationally, so it's better indexable/searchable etc?
Simple JSON, it is commonly used for apps instead of HTML because it's bit harder to manipulate HTML with all the nesteds.

> 3. How does it integrate with files?
Right now in this PR it's simple deploy file to directus files then use the /assets/<assetid> as preview url and stores { ... "assetId": <assetId> } in the data



> I think there is.
> WYSIWYG returns HTML that is easy to show on Android and iOS, app can simply fetch data and show it in a webview, while for editorjs, it needs to be server rendered to be shown on phones. If it's server rendered it also means that we must fetch data trough Directus, and not from db directly.

You could use the JSON just as regular text field, because it's stored in JSON field type, for rendering the field you need a rendering engine in both cases (HTML - webview or whatever browser you use, and for this interface output you could use your own, react/vue/ some native or even a console app)

> I think most apps should use WYSIWYG for blogs, and editorjs for page building.

Should or will?

Don't get me wrong, I am not against WYSIWYG, and I am not into merging this PR, just with PR it was more ease to reach and let people to test this if they want.

I am just pointing the fact that may be the WYSIWYG  is not used by all the users, but this doesn't means that it should be removed (at least at that moment)

rijkvanzanten:
I think this is better suited as a dedicated EditorJS extension instead of having it in the core codebase (we'll start stripping out more core things into extensions with the marketplace launching

PR #787, opened 2020-10-26
Assets fixes and improvements.

### LINKED ISSUE #474: Feature - Add asset option to avoid upscaling
Currently for assets presets we have only fit option which means to crop or to preserve aspect ratio. In some cases an option that prevent upscaling is required too.

As I checked in the next for resizing is used `sharp` package, which have `options.withoutEnlargement` option that control that feature.

### LINKED ISSUE #705: Custom asset presets don't work

rijkvanzanten:
- Assets improvements
- Use withoutEnlargement for more consistency
- Use withoutEnlargement for more consistency
- Fix item loading on first open
- Fix rendering assets with no settings
- Remove unused import
- Add migration for setting change

Fork of the original PR was deleted

Fixes #474, fixes #705


dimitrov-adrian:
Fixes https://github.com/directus/next/issues/791

PR #828, opened 2020-10-31
Small PWA touches

dimitrov-adrian:
Most notable

1. Optimize image files
2. Add splash screens for iOS
3. Set proper icon to use in Android splash screen
4. Use <%= BASE_URL %> in index.html instead of hard coded /admin
5. Move manifest file and set scope
6. Set theme color as current user's theme (mostly used in Android's Chrome)


rijkvanzanten (approved):

rijkvanzanten:
Thanks @dimitrov-adrian!

This is probably impossible, but it would be cool if we could come up with a way to dynamically change the icons/splash screens based on the saved project logo 🤔

dimitrov-adrian:
I thought about it, but then the manifest should be generated from the API, and the admin should upload by fields at least one svg logo icon for maskable, and then a lot of presets should exists only for the iOS splashes.

PR #860, opened 2020-11-03
Small CSS Update to public-view.vue for better mobile support

jekuer:
Add small fixes to optimize mobile support and scale elements propertly without overlapping each other or the screen.

This tiny update helps to keep elements in place for very small (and strange) screen resolutions. Since it is that tiny, I did not discuss it before the PR. Hope that's ok.

### DIFF
--- a/app/src/views/public/public-view.vue
+++ b/app/src/views/public/public-view.vue
@@ -133,6 +133,7 @@ export default defineComponent({
 		transition: max-width var(--medium) var(--transition);
 
 		.content {
+			max-width: 100%;
 			width: 340px;
 		}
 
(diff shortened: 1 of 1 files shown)


benhaynes:
Thank you @jekuer !! We'll check this out ASAP...

jekuer:
Maybe some details on the changes:

- Width 80% for the foreground element. At the moment it has an absolute max-width. On smaller screens, this leads to the element hitting the borders of the background element and even pushing it into the login form.
- Max-width of the title-box and the login form. At the moment it moves out of the view on smaller screens.

rijkvanzanten (approved):

Issue #871, opened 2020-11-04
CORS is not enabled by default


juancarlosjr97:
Steps to reproduce:

1. Create a new instance of directus `npx create-directus-project testing-directus-locally` and then `npx directus start`
2. Execute the command

```
fetch('http://localhost:8055/graphql', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    'Accept': 'application/json',
  },
  body: JSON.stringify({query: "query allSchemaTypes { __schema { types { name kind description } } }"})
})
  .then(r => r.json())
  .then(console.log);
```

3. The result is "rejected" and the expected result is "fulfilled"

4. Go to `.env` and add `CORS_ENABLED="true"`

5. Rerun `npx directus start`

6. Execute

```
fetch('http://localhost:8055/graphql', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    'Accept': 'application/json',
  },
  body: JSON.stringify({query: "query allSchemaTypes { __schema { types { name kind description } } }"})
})
  .then(r => r.json())
  .then(console.log);
```

and the result will be `fulfilled` as it should be expected.

------

CORS should be enabled by default which is not enabled by default

juancarlosjr97:
I just found out that is disabled by default, compared to the documentation that says is enabled

`node_modules/directus/dist/env.js` - `CORS_ENABLED: false,`

Either update documentation or enable CORS by default :)

rijkvanzanten:
Fixed in #872

PR #872, opened 2020-11-04
Updating CORS_ENABLED to true by default

juancarlosjr97:
Updating CORS_ENABLED to true by default as the documentation describes 

Reference: [directus/next] New comment on issue #871: CORS is not enabled by default

### DIFF
--- a/api/src/env.ts
+++ b/api/src/env.ts
@@ -29,7 +29,7 @@ const defaults: Record<string, any> = {
 
 	CORS_ENABLED: false,
 
-	CACHE_ENABLED: false,
+	CACHE_ENABLED: true,
 	CACHE_STORE: 'memory',
 	CACHE_TTL: '30m',
 	CACHE_NAMESPACE: 'system-cache',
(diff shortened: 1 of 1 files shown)


maartenvn:
You defaulted `CACHE_ENABLED` to true instead of `CORS_ENABLED`

api/src/env.ts
```diff
@@ -29,7 +29,7 @@ const defaults: Record<string, any> = {
 
 	CORS_ENABLED: false,
 
-	CACHE_ENABLED: false,
+	CACHE_ENABLED: true,
```
maartenvn, line 32: You meant to change `CORS_ENABLED` but you changed `CACHE_ENABLED` right?

maartenvn (changes requested):

api/src/env.ts
juancarlosjr97: Yes, tails of making a mistake half-asleep.... Thank you @maartenvn

rijkvanzanten (approved):

rijkvanzanten:
@fatihky that was resolved a commit later, see https://github.com/directus/next/pull/872/files for the final changes in this PR

fatihky:
@rijkvanzanten sorry, I see. thank you for the information. I was just looking at individual commits so I missed that.

PR #877, opened 2020-11-04
Use knex instance property in MetaService class

nickrum:
Just a small thing I noticed while working on #795. This should prevent issues if `MetaService` is ever used inside a transaction.


rijkvanzanten:
Good catch @nickrum, thanks!

rijkvanzanten (approved):

PR #878, opened 2020-11-04
Add tooltips to the icons in permissions header

nickrum:
This was in v8 and should help users to understand the meaning of those icons and their respective columns.


benhaynes:
Good catch! Thanks @nickrum ... we'll definitely need to do a clean sweep over the app for polishing tooltips.

nickrum:
I agree with that. Tooltips are certainly very helpful for any kind of user.

rijkvanzanten (approved):

rijkvanzanten (approved):

rijkvanzanten:
(Liked it so much, approved it twice! ʕ•ᴥ•ʔ )

PR #879, opened 2020-11-04
Use _and operator when converting filters to query

nickrum:
This fixes using multiple filters with the same field.

@rijkvanzanten what's your opinion on only allowing a single root field on filters as using `_and` essentially does the same thing without the limitation of not being able to specify the same field name multiple times?

Also, I think the corresponding readme file is a bit out of date. Shall I update it?

### DIFF
--- a/app/src/utils/filters-to-query/filters-to-query.ts
+++ b/app/src/utils/filters-to-query/filters-to-query.ts
@@ -1,8 +1,8 @@
 import { Filter } from '@/types/';
-import { set, clone } from 'lodash';
+import { clone } from 'lodash';
 
 export default function filtersToQuery(filters: readonly Filter[]) {
-	const filterQuery: Record<string, any> = {};
+	const filterList: Record<string, any>[] = [];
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
> what's your opinion on only allowing a single root field on filters

I would like to keep supporting

```js
{
  field: value,
  anotherField: anotherValue
}
```

as well, as it greatly improves DX in the REST API ( `?filter[field]=value&filter[anotherField]=anotherValue` )

> Also, I think the corresponding readme file is a bit out of date. Shall I update it?

Yes please!

rijkvanzanten:
Ps, @nickrum are you [on Discord](https://discord.gg/directus)? Makes talking through these kinds of things a little easier 🙂

rijkvanzanten (approved):

nickrum:
I see, thanks!

Yes, I do. Actually, I joined the Directus Discord a few days ago but haven't found the time to introduce myself 🙂

PR #880, opened 2020-11-04
Only enable save options if a value has changed

nickrum:
This feature was in v8 as well.

### DIFF
--- a/app/src/components/v-form/form-field.vue
+++ b/app/src/components/v-form/form-field.vue
@@ -21,7 +21,7 @@
 				:field="field"
 				:value="_value"
 				:initial-value="initialValue"
-				@input="$emit('input', $event)"
+				@input="emitValue($event)"
 				@unset="$emit('unset', $event)"
 				@edit-raw="showRaw = true"
 			/>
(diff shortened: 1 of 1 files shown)


benhaynes:
I believe the thinking here was that if you have default values set for fields in the form, you should be able to click save immediately since those are valid values.

This is different from v8, and I could go either way.

However, also keep in mind that some values are dynamic (eg `user_created` and `date_created`) and will save a custom value even without entering something into an input. Similarly, custom extensions may pull in external data on save, and may not be manually updated in the form either. Just food for thought.

nickrum:
Thanks for the feedback @benhaynes. I am very certain, I have not thought of all possible edge cases that may be affected by this change.

From a users perspective this should only affect editing existing items, so saving newly created items with default values should still work.

Currently items with dynamic values can only be edited and saved if another field has been changed. To change for example the `date_updated` field without changing any other value, a user has to change the value of a field and change it back to the original value. If purposely changing one of those dynamic values should still be possible, I think there should be a dedicated option for that.
I think this also applies to custom extensions, but I am not too sure about that.

benhaynes:
Ahh, for edits this makes perfect sense... and it should only show the save button as activated once you've made edits. Thanks!!

rijkvanzanten (approved):

rijkvanzanten:
@benhaynes The technical change in this PR is that it unsets the staged value if you edit and then undo your change

rijkvanzanten (approved):

nickrum:
Sorry, I should have been more specific in the first place!

PR #920, opened 2020-11-07
fix drawer on mobile

hamzahio:
the screen drawer on mobile is broken small fix in css please see images.

Screen Shot 2020-11-08 at 12 37 28 AM

### DIFF
--- a/app/src/components/v-drawer/v-drawer.vue
+++ b/app/src/components/v-drawer/v-drawer.vue
@@ -118,6 +118,9 @@ export default defineComponent({
 <style>
 body {
 	--v-drawer-max-width: 856px;
+	--v-drawer-cancel-top: 32px;
+	--v-drawer-cancel-left: -76px;
+
 }
 </style>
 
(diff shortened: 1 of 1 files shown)


benhaynes:
Thank you @hamzahio — this is a big help! We'll review ASAP...

rijkvanzanten:
Not sure how I feel about absolutely positioning that cancel button on top of the already existing menu button. I feel like I'd rather update the header bar to show a cancel button in case of the drawer (with a prop) and have that emit a cancel event so we don't end up with two buttons on top of eachother :thiknk

hamzahio:
@rijkvanzanten as I stated before it was quick fix without change any button behaviour.
I think if we change the header bar button will increase the complexity. what you think?

rijkvanzanten:
It will, and it won't. Doing it the more hacky way makes it harder to maintain in the future, as it's a "non standard" kinda fix on top of an existing structure

rijkvanzanten:
Closed in favor of https://github.com/directus/next/commit/0457509e9e6aaf8a4bf418ab09517cfa84b9faff

